PR #219, opened 2015-03-22
new string for storage folder (not translatable)

cbenkovic:


X4mp:
Please change the string Name to something a litte bit mir detailed. Like ext_storage_directory_name

PR #220, opened 2015-03-22
new string for storage folder (not translatable)

cbenkovic:


Paintroid/src/main/res/values/string.xml
```diff
@@ -21,6 +21,7 @@
 <resources xmlns:tools="http://schemas.android.com/tools">
 
     <string name="app_name">Pocket Paint</string>
+    <string name="ext_storage_directory_name" translatable="false">Pocket Paint</string>
```
X4mp: move this string down to the  section

btw, you can change your PR by simply pushing the same branch with an additional commit ;)
Also please squash your commits before pushing.

PR #225, opened 2015-10-22
solved resize width bug and added testChangeCroppingHeightAndCheckWidth

AlexWieland:


Paintroid/src/main/java/org/catrobat/paintroid/tools/implementation/ResizeTool.java
```diff
@@ -140,6 +141,7 @@ protected void drawToolSpecifics(Canvas canvas) {
 				mBoxHeight = mBoxWidth;
 				mBoxWidth = tempHeight;
 			}
+			mBoxWidth = tempWidth;
```
X4mp: see line 142. Is it necessary to save tempWidth back to mBoxWidth 2 times?

PR #231, opened 2016-03-23
TextTool and test fixex for Android 6

cbenkovic:


Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/TextToolCommand.java
```diff
@@ -0,0 +1,80 @@
...
+		mBoxWidth = boxWidth;
+		mBoxHeight = boxHeight;
+		mToolPosition = toolPosition;
+		mRotationAngle = rotationAngle;
+	}
+
+	@Override
+	public void run(Canvas canvas, Bitmap bitmap) {
+		notifyStatus(NOTIFY_STATES.COMMAND_STARTED);
+
```
X4mp: add parameter check (bitmap == NULL).
see Stamp Command implementation (StampCommand.java:64)

Paintroid/src/main/java/org/catrobat/paintroid/tools/ToolFactory.java
```diff
@@ -68,6 +69,8 @@ public static Tool createTool(Activity context, ToolType toolType) {
 			return new MoveZoomTool(context, toolType);
 		case ROTATE:
 			return new RotationTool(context, toolType);
+		case TEXT:
+				return new TextTool(context, toolType);
```
X4mp: intentation

PR #241, opened 2016-04-06
gradle build version, translations update

cbenkovic:


X4mp:
Why did you change all the language files? You should only add new Strings
to the default Strings.xml

cbenkovic:
because these are the updates from crowdin

PR #249, opened 2016-05-27
add tolerance to fill tool

cbenkovic:


Paintroid/src/main/java/org/catrobat/paintroid/tools/helper/FillAlgorithm.java
```diff
@@ -89,7 +92,7 @@ private Range generateRangeAndReplaceColor(int row, int col, boolean direction)
 
 		int i;
 		for (i = col; i >= 0; i--) {
-			if (mPixels[row][i] == mReplacementColor) {
+			if (mPixels[row][i] == mReplacementColor || isPixelWithinColorTolerance(mPixels[row][i])) {
```
wslany: Could the `mPixels[row][i] == mReplacementColor || ` be removed here as well as in lines 106 and 122 below, so that this line becomes `if (isPixelWithinColorTolerance(mPixels[row][i])) {` ?

AlxO: Am 28.05.2016 um 08:16 schrieb Wolfgang Slany:
> Could the `mPixels[row][i] == mReplacementColor || ` be removed here as well as in lines 106 and 122 below, so that this line becomes `if (isPixelWithinColorTolerance(mPixels[row][i])) {` ?

that should be an performance improvement. if it is the exactly the same
color the slower isPixelWithinColorTolerance() function should not be
called.

Paintroid/src/main/java/org/catrobat/paintroid/tools/helper/FillAlgorithm.java
```diff
@@ -136,4 +139,13 @@ private void checkRangeAndGenerateNewRanges(Range range, int row, boolean direct
 		}
 	}
 
+	private boolean isPixelWithinColorTolerance(int pixel) {
```
wslany: Since this method is called very often, I propose to make it a bit faster, as follows: Replace `mColorToleranceThreshold` by `mColorToleranceThresholdSquared` to avoid having to call the cpu intensive `Math.sqrt` method for each pixel. 

Additionally, to make it even more efficient, avoid the intermediate `diff` variable.

The code thus could look as follows:

	private boolean isPixelWithinColorTolerance(int pixel) {
		return ColorToleranceThresholdSquared >=
                       Math.pow(Color.red(pixel) - Color.red(mReplacementColor), 2)
			+ Math.pow(Color.green(pixel) - Color.green(mReplacementColor), 2)
			+ Math.pow(Color.blue(pixel) - Color.blue(mReplacementColor), 2)
			+ Math.pow(Color.alpha(pixel) - Color.alpha(mReplacementColor), 2);
	}

I suggest to add a comment that explains the reason for this speed optimization, so that it does not get undone during code review.

`mColorToleranceThresholdSquared` needs to be computed only once per fill command in line 58, as shown above.

Paintroid/src/main/java/org/catrobat/paintroid/tools/helper/FillAlgorithm.java
```diff
@@ -53,6 +55,7 @@ public FillAlgorithm(Bitmap bitmap, Point clickedPixel, int targetColor, int rep
 		mTargetColor = targetColor;
 		mReplacementColor = replacementColor;
 		mRanges = new LinkedList<Range>();
+		mColorToleranceThreshold = colorToleranceThreshold;
```
wslany: As explained below in my comment at line 142, replace by

		mColorToleranceThresholdSquared = Math.pow(colorToleranceThreshold, 2);

wslany:
Added a few in-line comments to increase speed.

Paintroid/src/main/java/org/catrobat/paintroid/tools/helper/FillAlgorithm.java
wslany: Ahh, OK, you are right. I was not thinking correctly ;-)

wslany:
Can lines 51-53 in _FillAlgorithm.java_ be eliminated? They seem to serve no purpose. I mean

		for (int i = 0; i < mHeight; i++) {
			mBitmap.getPixels(mPixels[i], 0, mWidth, 0, i, mWidth, 1);
		}

AlxO:
Am 28.05.2016 um 09:17 schrieb Wolfgang Slany:
> Can lines 51-53 in _FillAlgorithm.java_ be eliminated? They seem to serve no purpose. I mean

it initializes mPixels. maybe there is an other way to do it.

wslany:
> it initializes mPixels. maybe there is an other way to do it.

You are right, and I understand now. Sorry for your time! I'll try to be more careful with my comments from now on.

PR #347, opened 2017-04-10
PAINT-152 Multilingual Feature on Application Level

ndrnour:
The app must provide an interface that handles user input to switch languages independent of the smartphone’s global settings. This feature should list all languages supported by the app including languages not yet supported by the operating system.


robertpainsi:
@Catrobat/paintroid, please review again.

PR #350, opened 2017-04-12
Removes AllIntegrationTests.

mfuchs:
AllIntegrationTests inhibts the parallelisation of tests.

Background info
---------------
In our case the parallelisation of tests in Jenkins works as follows:

1) A build needs to run successfully.
   Assume that it contained just the tests
   * BrushPickerIntegrationTest (short Brush)
   * ColorDialogIntegrationTest (short Color)
   * AllIntegrationTests (short All), which contains the above tests
2) For now we configured Jenkins to split all tests into two batches
   that take equal time.
   Obviously time(All) = time(Brush) + time(Color)
   Thus batch1 = [All] and batch2 = [Brush, Color]
3) Instead of directly running the tests specified in batch1 Jenkins
   excludes all _other_ tests.
   So when running batch1 Jenkins actually excludes every other batch
   and thus excludes Brush and Color.
4) Of course the build will fail since to compile All needs Brush and Color.


mfuchs:
test this please

PR #367, opened 2017-05-09
PAINT-172 Localized Font Faces for Right-to-Left Languages

ndrnour:
Localized product might have various UI fonts’ types than the original products. Local users will use the best font type according to the font capabilities of the particular target language.


Paintroid/src/androidTest/java/org/catrobat/paintroid/test/junit/intro/WelcomeActivity_RTL_Test.java
```diff
@@ -0,0 +1,120 @@
...
+                R.layout.islide_landscape,           3
+                R.layout.islide_getstarted};         4
+        */
+        assertEquals(0, viewPager.getCurrentItem());
+        swipeToRight();
+        assertEquals(1, viewPager.getCurrentItem());
+        swipeToRight();
+        assertEquals(2, viewPager.getCurrentItem());
+        solo.sleep(400);
+        solo.clickOnText(solo.getString(R.string.next));
```
duchuyctlk: IMO `solo.getString(R.string.next)` will always return a same constant string, so it could be extracted in to a variable to avoid multiple method calls.

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/junit/intro/WelcomeActivity_RTL_Test.java
```diff
@@ -0,0 +1,120 @@
...
+        solo.sleep(600);
+        assertEquals(4, viewPager.getCurrentItem());
+        //back to start to test nextButton
+        swipeToLeft();
+        swipeToLeft();
+        swipeToLeft();
+        swipeToLeft();
+        assertEquals(0, viewPager.getCurrentItem());
+        solo.clickOnText(solo.getString(R.string.next));
+        solo.sleep(400);
```
duchuyctlk: Integer constant for sleeping could be extracted too.

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/junit/intro/WelcomeActivity_RTL_Test.java
ndrnour: @duchuyctlk  Thanks for your review, could you please explain me why I should make that? Also do you think it will make a difference!!

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/junit/intro/WelcomeActivity_RTL_Test.java
ndrnour: @duchuyctlk  could I say why ?

duchuyctlk: Just in case you want the amount of sleeping time remains intentionally the same, not only coincidentally the same. You will save effort later when you want to raise that intentional value.

ndrnour:
hey!
I cant access to have a look at the conflicts! how can I resolve them ?

ndrnour:
could someone please review this PR?

PR #375, opened 2017-05-30
Paint 128 layer adaptions bugs

AlexWieland:


cbenkovic:
bug: move a layer outside the menu from top to bottom (drag in the layer menu from below). The bottommost layer disappears downwards which leads to an unstable behaviour of the layer menu.

Paintroid/src/main/java/org/catrobat/paintroid/ui/dragndrop/BrickDragAndDropLayerMenu.java
```diff
@@ -1,34 +1,91 @@
 package org.catrobat.paintroid.ui.dragndrop;
```
cbenkovic: add ...Copyright (C)..., also in the other dragndrop files

PR #383, opened 2017-08-09
Paint-175 color chooser after eraser

ThomasSchwengler:
Fixed the problem that text would invisible after using the eraser.


ThomasSchwengler:
Testing can be done after merging #385.

PR #384, opened 2017-08-12
PAINT-176: Pipette blending color from all layers

ThomasSchwengler:


ThomasSchwengler:
Added an espresso integration test, since the original unit test doesn't work in this case.

I'm unsure on how to handle this, should I remove the original unit tests and recreate them in espresso?

PR #425, opened 2017-09-13
PAINT-209 Delete old Layer Commands to fix Layer Memory Leak

AlexWieland:


cbenkovic:
Please fix the crash when redo ing a layer command on a different layer

Steps for reproducing
# add two layers
# undo once
# draw something -> the redo list for this layer is cleared
# change to the other layer
# here you can still redo the layer command, which leads to the crash

PR #471, opened 2018-01-15
PAINT-271 replace Alarabiya Font with STC Font

ndrnour:


ndrnour:
please run espresso tests

PR #482, opened 2018-01-31
PAINT-255: language update

chbinder:


ThomasSchwengler:
Please replace the ... characters with … (U+2026)
You can just do a search-replace if it doesn't work via crowding.

PR #493, opened 2018-03-14
Master

abhikhrt:


ThomasSchwengler:
Hello! What's the reason why you want this merged?
We are using [GitFlow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow), which means that the master branch will only get commits from a release branch, not directly from develop.

PR #494, opened 2018-03-19
Paint-295: Restructure packages

ThomasSchwengler:
Move packages to corresponding catroid namespace


daberni:
Are you aware, that all homescreen-shortcuts will be removed from devices when updating to the new app version because the launcher-activity class name changed?

adding an alias in the manifest would be a workaround for this issue

ThomasSchwengler:
That's probably a good idea, I will keep that in mind

PR #505, opened 2018-05-09
PAINT-298: Update dependencies and fix ndk issues

ThomasSchwengler:
* Filter abi support to all supported abis by ndk `>= 17`
* Update dependencies
* ~Upgrade `buildToolsVersion`~
* Upgrade `compileSdkVersion` to 26
* Upgrade gradle wrapper
* Remove custom runner, this shouldn't be needed anymore with
  the new android emulator on jenkins
* Migrate dependency configuration to 3.0.0 ([ref](https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration))
* Fix `Jenkinsfile` apk location


build.gradle
```diff
@@ -21,18 +21,17 @@
...
     }
 }
 
 allprojects {
     repositories {
         jcenter()
-        maven {
-            url "https://maven.google.com"
-        }
+        maven { url 'https://maven.google.com' }
```
daberni: could be `google()` too

Paintroid/build.gradle
```diff
@@ -41,18 +41,19 @@ if (project.hasProperty('independent')) {
 }
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion '25.0.3'
+    compileSdkVersion 26
+    buildToolsVersion '26.0.2'
```
daberni: buildToolsVersion is ignored with android-gradle-plugin versions >= 3
this line can be removed

build.gradle
```diff
@@ -21,18 +21,17 @@
 buildscript {
     repositories {
         jcenter()
+        google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.3'
+        classpath 'com.android.tools.build:gradle:3.0.1'
```
redeamer: If the gradle version is upgraded, the output of the debug *.apk changes. There will be an additional subdirectory under `build/outputs/apk/` for the build flavor (eg debug). Please respect the new Location in the `Jenkinsfile`. It is defined inside the `APK_LOCATION_DEBUG` variable. It should change from `.../apk/Paintroid-debug.apk` to `.../apk/debug/Paintroid-debug.apk`

ThomasSchwengler: Should be fixed now

Paintroid/build.gradle
ThomasSchwengler: Thanks, removed the `buildToolsVersion` line

build.gradle
ThomasSchwengler: You are right, switched it to `google()`

PR #528, opened 2018-08-02
JENKINS-229 Also track code coverage for unit tests.

mfuchs:
* Code coverage is collected for all tests that run on the emulator
  and also for unit tests.
* Before builds there is a clean-up step.
  This step ensures that the coverage reports are removed before the build.
  Otherwise PMD would diagnose that these xml files do not follow its guidelines.
* To make the cleanup step work files avds are kept.
* Uses a newer JaCoCo version.

Compare [old](https://jenkins.catrob.at/job/Paintroid/job/develop/177/cobertura/org_catrobat_paintroid_listener/DrawingSurfaceListener_java/) vs [new](https://jenkins.catrob.at/job/lab/job/Matthias_Fuchs/job/CoveragePaintroid/2/cobertura/org_catrobat_paintroid_listener/DrawingSurfaceListener_java/).


mfuchs:
@redeamer can please also review these changes?

Jenkinsfile
```diff
@@ -46,19 +46,34 @@ pipeline {
...
 	}
 
 	stages {
+		stage('Cleanup') {
+			steps {
+				// Clean all files and directories not tracked by git and that are not part of .gitignore.
+				// This includes the code coverage files that could otherwise lead to false positives
+				// during static analysis.
+				// Note: The build directory is not cleaned, this is done by gradle.
+				sh 'git clean -fd'
```
redeamer: Jenkins (resp the job) is configured to do a 'git clean -fdx' ('Clean before checkout'-option) on the used repository right before running the job. Therefore this stage 'duplicates' that task and would be a no-op.

.gitignore
```diff
@@ -27,3 +27,10 @@ build/
 
 # Cache for build scripts
 __pycache__
+
+# Files generated for the Android Emulator.
```
redeamer: Is this important just for the 'Cleanup'-stage? If yes, that would not make a difference, as Jenkins will call a 'git clean -fdx' ('Clean before checkout'-option) before running the job, so those files would be removed anyhow. Which removes the need to cover requirements like: resetting the emulator state, reconfigure the emulator if the emulator config is changed, ...
Is it intended for the developer machine? Then it would be fine I think (can find pro's and con's for that).

Jenkinsfile
```diff
@@ -46,19 +46,34 @@ pipeline {
 		ANDROID_NDK = ""
 
 		PYTHONUNBUFFERED = "true"
+
 		JACOCO_XML = "Paintroid/build/reports/coverage/debug/report.xml"
+		JACOCO_UNIT_XML = "Paintroid/build/reports/jacoco/jacocoTestDebugUnitTestReport/jacocoTestDebugUnitTestReport.xml"
```
redeamer: I had the intention to replace all 'module'-occurrences with the variable GRADLE_PROJECT_MODULE_NAME so the Jenkinsfiles are more easily comparable between Paintroid and Catroid.
For consistency's sake we should either replace Paintroid with ${env.GRADLE_PROJECT_MODULE_NAME} or vice-versa and ditch the variable entirely. If we go for the latter option, I would suggest this could be another commit/PR which takes the restructured environment section from Catroid as template, because it includes a slightly improved documentation for the developer about the environment settings defined here.

buildScripts/build_helper_run_unit_tests
```diff
@@ -22,7 +22,7 @@ The environment variable ANDROID_SDK_ROOT needs to be set.""")
 build_helper_functions.check_number_of_parameters(valid_param_count_min=0, valid_param_count_max=1, usage_func=usage)
 
 ## RUN tests
-test_runner_cmd = [ build_helper_functions.get_relative_gradle_name(), '-Pjenkins', 'clean', 'testDebug' ]
+test_runner_cmd = [ build_helper_functions.get_relative_gradle_name(), '-Pjenkins', 'clean', 'jacocoTestDebugUnitTestReport' ]
 ## pass filter if argument is given
 if len(sys.argv) > 1:
```
redeamer: jacocoTestDebugUnitTestReport does not seem to provide the '--tests' filter. Therefore the optional argument can be removed. As consequence 'build_step_run_unit_tests__all_tests' as a wrapper with defined arguments to 'build_helper_run_unit_tests' would be useless and they could be merged in an 'build_step_run_unit_tests'.

Jenkinsfile
```diff
@@ -95,13 +114,13 @@ pipeline {
...
 				// See also JENKINS-212 on jira.catrob.at
-				sh "if [ -e '$JACOCO_XML' ]; then ./buildScripts/cover2cover.py $JACOCO_XML > $COBERTURA_XML; fi"
+				sh "if [ -e '$JACOCO_XML' ]; then ./buildScripts/cover2cover.py $JACOCO_XML > $JAVA_SRC/coverage2.xml; fi"
 			}
 
 			post {
 				always {
 					junit '**/*TEST*.xml'
-					step([$class: 'CoberturaPublisher', autoUpdateHealth: false, autoUpdateStability: false, coberturaReportFile: COBERTURA_XML, failUnhealthy: false, failUnstable: false, maxNumberOfBuilds: 0, onlyStable: false, sourceEncoding: 'ASCII', zoomCoverageChart: false, failNoReports: false])
+					step([$class: 'CoberturaPublisher', autoUpdateHealth: false, autoUpdateStability: false, coberturaReportFile: "$JAVA_SRC/coverage*.xml", failUnhealthy: false, failUnstable: false, maxNumberOfBuilds: 0, onlyStable: false, sourceEncoding: 'ASCII', zoomCoverageChart: false, failNoReports: false])
```
redeamer: If cleaning is still an issue as Jenkins runs 'git clean -fdx' before every build, I could imagine that we simply remove the two explicitly created report files right after the are 'published'. I have not tried if that would work correctly, but it would IMHO have the least impact if modifications of that stage are removed in the stage again.

Jenkinsfile
```diff
@@ -46,19 +46,34 @@ pipeline {
...
-		COBERTURA_XML = "Paintroid/src/main/java/coverage.xml"
+		JAVA_SRC = "Paintroid/src/main/java"
+
 	}
 
 	options {
 		timeout(time: 2, unit: 'HOURS')
 		timestamps()
-                buildDiscarder(logRotator(numToKeepStr: '30'))
+		buildDiscarder(logRotator(numToKeepStr: '30'))
```
redeamer: NO sarcasm: what a relief, thanks for fixing the inconsistency!

.gitignore
mfuchs: I missed the "Clean before checkout" setting.
So apparently the emulator is always recreated.
I'll remove this part.

Jenkinsfile
mfuchs: Vim points out to me if there is an inconsistency. ;)

Jenkinsfile
mfuchs: In the job I tried this with I did not call the cleanup step of git.
And when checking the config in the GUI I missed it.
I'll remove these things.

buildScripts/build_helper_run_unit_tests
mfuchs: Will do.

buildScripts/build_step_run_unit_tests__all_tests
```diff
@@ -17,11 +17,11 @@ allows to add a test-filter, but currently all tests are run.
```
redeamer: With the changed functionality of the script, the usage is now outdated.

PR #539, opened 2018-09-04
JENKINS-215 Create, start and stop emulator via gradle.

mfuchs:
* Three new adb_tasks were added: startEmulator, stopEmulator, and retrieveLogcat
* startEmulator creates an avd if necessary and the starts the emulator.
  An already running emulator is reused.
* retrieveLogcat downloads the logcat.txt file from the device.
* The buildScripts are kept around until the Android SDK can be
  installed via gradle, see JENKINS-217.
* The Jenkinsfile uses the gradle tasks directly.
  This makes it more transparent of what is happening.
  It is the task's job to work correctly on both Jenkins and locally.
* Much of the groovy code is type checked during compilation.
  As a result not all tasks need to be run to find errors in buildSrc.
  This can also be seen as preparation to move buildSrc into its
  own gradle plugin.

Note: Does currently not support running multiple emulators at the same time.
      The avd name is not used to look up the pid of a running emulator.
      Of course when you use docker you can run multiple instances at once.

Note2: This is an initial PR. PRs in the following weeks will add further steps like installing the SDK or handling multiple images. It is here to see if the solution is feasible per se.
The changes were tested on both Linux and Windows though.

@redeamer please also check the changes.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -83,7 +83,7 @@ pipeline {
 
 		stage('Static Analysis') {
 			steps {
-				sh "./buildScripts/build_step_run_static_analysis"
+				sh './gradlew clean pmd checkstyle lint'
 			}
 
 			post {
@@ -98,7 +98,7 @@ pipeline {
 		stage('Unit and Device tests') {
 			steps {
 				// Run local unit tests
-				sh "./buildScripts/build_step_run_unit_tests__all_tests"
+				sh './gradlew -Pjenkins clean jacocoTestDebugUnitTestReport'
 				// Convert the JaCoCo coverate to the Cobertura XML file format.
 				// This is done since the Jenkins JaCoCo plugin does not work well.
 				// See also JENKINS-212 on jira.catrob.at
@@ -107,8 +107,11 @@ pipeline {
 				sh "mv ${env.GRADLE_PROJECT_MODULE_NAME}/build ${env.GRADLE_PROJECT_MODULE_NAME}/build-unittest"
 
 				// Run device tests
-				sh "./buildScripts/build_step_run_tests_on_emulator__all_tests"
-
+				sh '''
+					./gradlew startEmulator adbDisableAnimationsGlobally
+					./gradlew -Pjenkins clean createDebugCoverageReport || true
+					./gradlew adbResetAnimationsGlobally retrieveLogcat
+				'''
 				// Convert the JaCoCo coverate to the Cobertura XML file format.
 				// This is done since the Jenkins JaCoCo plugin does not work well.
 				// See also JENKINS-212 on jira.catrob.at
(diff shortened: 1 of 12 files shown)


buildSrc/src/main/groovy/org/catrobat/gradle/Exceptions.groovy
```diff
@@ -0,0 +1,48 @@
...
+}
+
+@TypeChecked
+@InheritConstructors
+class NoAvdException extends AndroidResourceException {
+}
+
+@TypeChecked
+@InheritConstructors
+class NoDeviceExcpetion extends AndroidResourceException {
```
ThomasSchwengler: please fix the typo

buildSrc/src/main/groovy/org/catrobat/gradle/Exceptions.groovy
```diff
@@ -0,0 +1,48 @@
...
+}
+
+@TypeChecked
+@InheritConstructors
+class DeviceNotFoundException extends NoDeviceExcpetion {
+}
+
+@TypeChecked
+@InheritConstructors
+class BootIncompleteExcpetion extends AndroidResourceException {
```
ThomasSchwengler: please fix the typo

Jenkinsfile
```diff
@@ -120,15 +123,15 @@ pipeline {
...
+					sh './gradlew stopEmulator'
+					archiveArtifacts 'logcat.txt'
 				}
 			}
 		}
 
 		stage('Build Debug-APK') {
 			steps {
-				sh "./buildScripts/build_step_create_debug_apk"
+				sh './gradlew clean assembleAndroidTest'
```
ThomasSchwengler: This might need to be changed to 'assembleDebug' after #540 is merged, (or in #540) if this gets merged first.

mfuchs:
I fixed the typos as a separate commit. :)
I can squash the commit in the end if you want. Though that is also possible directly via github.

 :exclamation: I just realised that I am still not part of the Paintroid team. Thus my PRs still use the Jenkinsfile of develop for security reasons.
I created my own jobs in labs though during development.

mfuchs:
I have more rights now and will start start the build-job myself.
Please do not start the job yourself, as I want to check some things first. :)

Jenkinsfile
ThomasSchwengler: #540 is merged, please change this to "assembleDebug"

mfuchs:
:information_source: if #541 is merged before adapt the Jenkinsfile to also provide -PenableCoverage

ThomasSchwengler:
#541 got merged, please adapt the Jenkinsfile if needed.

mfuchs:
We should probably wait until the Paintroid release and when the release branch was merged into develop.

mfuchs:
With the progress I made it is probably easier to skip this one and handle it directly with JENKINS-217.
I am keeping it open for now though to remember the steps I need to perform for #541

mfuchs:
I created #549 that covers all my changes so far.
If you want to have smaller commits I can also keep this PR and make #549 depend on this.

PR #544, opened 2018-09-12
PAINT-322 request permission at runtime if needed

SebastianGrief:


Paintroid/src/main/java/org/catrobat/paintroid/dialog/PermissionInfoDialog.java
```diff
@@ -0,0 +1,95 @@
+package org.catrobat.paintroid.dialog;
```
ThomasSchwengler: Please add copyright above

Paintroid/src/test/java/org/catrobat/paintroid/test/presenter/MainActivityPresenterTest.java
```diff
@@ -1001,12 +1006,25 @@ public void testOnSaveImagePreExecuteThenShowProgressDialog() {
 	}
 
 	@Test
-	public void testOnSaveImagePostExecuteWhenFailedThenShowDialog() {
+	public void testOnSaveImagePostExecuteWhenFailedAndPermissionGrantedThenShowDialog() {
+		doReturn(true).when(navigator).doIHavePermission(Manifest.permission.WRITE_EXTERNAL_STORAGE);
+
```
ThomasSchwengler: `doReturn(true).when(navigator).isSdkAboveOrEqualM();` is probably missing here

ThomasSchwengler:
Regarding the behavior: This will only show the permission dialogs once saving fails. 

The ticket on jira states as expected behavior:

> The user is asked for the `WRITE_EXTERNAL_STORAGE` permission at runtime
> with an explanation why the premission is needed _before_ 
> trying to write to external storage _if_ the premission is not already granted.

ThomasSchwengler:
Other than that it seems fine :+1:

SebastianGrief:
> Regarding the behavior: This will only show the permission dialogs once saving fails.
> 
> The ticket on jira states as expected behavior:
> 
> > The user is asked for the `WRITE_EXTERNAL_STORAGE` permission at runtime
> > with an explanation why the premission is needed _before_
> > trying to write to external storage _if_ the premission is not already granted.

Should it be checked at startup?

ThomasSchwengler:
> Should it be checked at startup?

No, it should be checked every time before a user tries to save something (The user might retract the permission while the app is running)

Paintroid/src/test/java/org/catrobat/paintroid/test/presenter/MainActivityPresenterTest.java
```diff
@@ -103,6 +114,7 @@
 	@Mock
 	private MainActivityContracts.BottomBarViewHolder bottomBarViewHolder;
 
+	@Spy
```
ThomasSchwengler: Presenter should be a regular instance, not a spy. Is this somehow needed?

SebastianGrief: Yes. 
It's needed for the 4 tests which verify that the previous action is repeated after the permission was granted. 1035-1103 
I can try to find another way to test that if needed.

Paintroid/src/test/java/org/catrobat/paintroid/test/presenter/MainActivityPresenterTest.java
```diff
@@ -1007,6 +956,154 @@ public void testOnSaveImagePostExecuteWhenFailedThenShowDialog() {
...
+	public void testSaveImageClickedWhenPermissionNotGrantedThenAskForPermissionShowInfoDialogIfNotGranted() {
+		doReturn(false).when(navigator).doIHavePermission(Manifest.permission.WRITE_EXTERNAL_STORAGE);
+		doReturn(true).when(navigator).isSdkAboveOrEqualM();
+
+		presenter.saveImageClicked();
+
+		verify(navigator).askForPermission(new String[] {Manifest.permission.WRITE_EXTERNAL_STORAGE},
+				PERMISSION_EXTERNAL_STORAGE_SAVE);
+
+		presenter.handlePermissionRequestResults(PERMISSION_EXTERNAL_STORAGE_SAVE,
```
ThomasSchwengler: Please stick to the `AAA` pattern for testing.

The `assert` phase should not be succeeded with additionals calls to the presenter.
These tests can be split up in at least two test cases each.

Paintroid/src/test/java/org/catrobat/paintroid/test/presenter/MainActivityPresenterTest.java
SebastianGrief: :980
:1011
:1042
:1075

After splitting.

Paintroid/src/test/java/org/catrobat/paintroid/test/presenter/MainActivityPresenterTest.java
```diff
@@ -1007,6 +955,132 @@ public void testOnSaveImagePostExecuteWhenFailedThenShowDialog() {
...
+				PERMISSION_EXTERNAL_STORAGE_SAVE_COPY);
+	}
+
+	@Test
+	public void testHandlePermissionRequestResultForRequestAfterSaveCopyClickedPermissionGranted() {
+		presenter.handlePermissionRequestResults(PERMISSION_EXTERNAL_STORAGE_SAVE_COPY,
+				new String[] {Manifest.permission.WRITE_EXTERNAL_STORAGE},
+				new int[] {PackageManager.PERMISSION_GRANTED});
+
+		verify(presenter).saveCopyClicked();
```
ThomasSchwengler: Regarding `@spy` - I see now.
Please don't use mockito to verify that the presenter called an internal function.
You should check that the presenter did the right things to the collaborating classes, in this case you should verify that the presenter calls `interactor.saveCopy` does _not_ call `navigator.askForPermission()` again.
Don't mock/spy the class under test

SebastianGrief: Okay

Paintroid/src/main/java/org/catrobat/paintroid/presenter/MainActivityPresenter.java
```diff
@@ -239,6 +272,46 @@ public void handleActivityResult(@ActivityRequestCode int requestCode, int resul
 		}
 	}
 
+	@Override
+	public void handlePermissionRequestResults(int requestCode, String[] permissions, int[] grantResults) {
+		switch (requestCode) {
```
ThomasSchwengler: Please simplify.
If the permission is not granted you can call
```
navigator.showPermissionDialog(PermissionInfoDialog.PermissionType.EXTERNAL_STORAGE, EXTERNAL_STORAGE_PERMISSION_DIALOG, requestCode);
```
If the permission is granted, switch over the cases and call the corresponding functions.

Paintroid/src/main/java/org/catrobat/paintroid/common/Constants.java
```diff
@@ -42,6 +42,14 @@
 	public static final String SAVE_QUESTION_FRAGMENT_TAG = "savebeforequitfragment";
 	public static final String CHOOSE_IMAGE_FRAGMENT_TAG = "chooseimagefragment";
 
+	public static final String EXTERNAL_STORAGE_PERMISSION_DIALOG = "externalstoragepermissiondialog";
+
+	public static final int PERMISSION_EXTERNAL_STORAGE = 0;
```
ThomasSchwengler: please move the `PERMISSION_*` ints to `MainActivityConstants` and group them in an `@interface` (Similar to `CreateFileRequestCode`).
You can then annotate the `requestCode` argument in functions like `handlePermissionRequestResults` with `@PermissionRequestCode` (or how ever you name the interface)

PR #549, opened 2018-09-17
JENKINS-215, JENKINS-217 Download dependencies, create, start and sto…

mfuchs:
…p emulator via gradle.

* Adds an own gradle plugin to handle interaction with the emulator.
* The android sdk/ndk can be installed when calling -PinstallSdk.
  This is done in the configuration step, to ensure that the plugins
  have all the information they need.
* Three new adb_tasks were added: startEmulator, stopEmulator, and retrieveLogcat
* startEmulator creates an avd if necessary and the starts the emulator.
  An already running emulator is reused.
  Use -Pemulator or the environment variable EMULATOR_OVERRIDE to specify the
  emulator to use.
* The emulator as well as the dependencies can be specified in a DSL in
  the build.gradle script.
  Multiple emulators can be added via emulator, furthermore templates can
  be specified via emulatorTemplate to share commong parts of emulators.
* Changing the specification of an emulator leads to the emulator being recreated.
* retrieveLogcat downloads the logcat.txt file from the device.
* The buildScripts are kept for now in cases issues arise.
* The Jenkinsfile uses the gradle tasks directly.
  This makes it more transparent of what is happening.
  It is the task's job to work correctly on both Jenkins and locally.
* Much of the groovy code is type checked during compilation.
  As a result not all tasks need to be run to find errors in buildSrc.
  This can also be seen as preparation to move buildSrc into its
  own gradle plugin that then lives in its own repo.

Note: Does currently not support running multiple emulators at the same time.
      The avd name is not used to look up the pid of a running emulator.
      Of course when you use docker you can run multiple instances at once.

### DIFF
--- a/Paintroid/build.gradle
+++ b/Paintroid/build.gradle
@@ -17,6 +17,34 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import org.catrobat.gradle.EmulatorsPlugin
+
+apply plugin: EmulatorsPlugin
+
+emulators {
+    dependencies {
+        sdk()
+        ndk()
+    }
+
+    emulator 'android24', {
+        avd {
+            systemImage = 'system-images;android-24;default;x86_64'
+            sdcardSizeMb = 200
+            hardwareProperties += ['hw.ramSize': 800, 'vm.heapSize': 128]
+            screenDensity = 'xhdpi'
+        }
+
+        parameters {
+            resolution = '768x1280'
+            language = 'en'
+            country = 'US'
+        }
+    }
+
+    shouldInstallEverythingAbove(project.hasProperty('installSdk'))
+}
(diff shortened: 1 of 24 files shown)


mfuchs:
Sorry for the huge code-dump.
JENKINS_215 took very long and my code evolved quite a lot in the meanwhile.

Paintroid/build.gradle
```diff
@@ -17,6 +17,35 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import org.catrobat.gradle.Installer
```
ThomasSchwengler: Please remove unnecessary import

mfuchs:
Squashed them all.

ThomasSchwengler:
:+1:

PR #552, opened 2018-09-19
PAINT-286: shape tool modifications

chbinder:


Paintroid/src/main/res/layout/dialog_pocketpaint_shapes.xml
```diff
@@ -17,8 +17,98 @@
...
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:weightSum="5"
+        android:orientation="horizontal" >
+
+        <SeekBar
+            android:id="@+id/pocketpaint_shape_stroke_width_seek_bar"
+            android:progressTint="@color/pocketpaint_colorAccent"
```
ThomasSchwengler: I think you can just remove this, the generated `AppCompatSeekBar` should use `colorAccent` automatically.

Paintroid/src/main/res/layout/dialog_pocketpaint_shapes.xml
```diff
@@ -17,8 +17,98 @@
...
+            android:thumbTint="@color/pocketpaint_colorAccent"
```
ThomasSchwengler: Same here.

PR #569, opened 2018-10-14
JENKINS-216 Uses the published androidemulators gradle plugin instead of the local one.

mfuchs:
The code to manage android emulators was moved into its own plugin.
Thus there is no reason anymore to keep both the python code (buildScripts)
and the gradle code (buildSrc) around.

The published version does not have a retrieveLogcat command.
Instead it always retrieves the logcat file automatically.

Note: The log parser rules as well as the cover2cover script are still needed.


mfuchs:
@ThomasSchwengler I am not sure how to avoid the dependencies block in the adb_tasks.gradle file.
I think it is both ugly and makes it easier to wrongfully have different dependencies in build.gradle and adb_tasks.gradle

We could use a global variable to share the version, though then we'd have to move the ext block up, to be usable by the dependencies.

ThomasSchwengler:
Please rebase.
You could just move the version to the `ext` block I guess, if that's possible. Otherwise it's a unavoidable evil I suppose. :)

mfuchs:
I was not able to access the variable in the root build.gradle itself when it is part of ext.
Just in case I tried to move the ext block above the dependencies.
Even when I just defined a variable outside of all blocks I was not able to access it in dependencies.

I guess that the closure used for dependencies is configured (the delegate) so that it only finds variables/functions within the ScriptHandler class.

PR #577, opened 2018-11-04
PAINTROID-10, JENKINS-251 Moves many responsibilities to the Dockerfile.

mfuchs:
* Install Android SDK/NDK and packages in Docker image

  This makes clear what dependencies are used.
  When new SDK/NDK versions are released the Paintroid developers can
  decide when to migrate to the newer versions.
  This is especially useful when newer versions cause issues,
  like the last NDK release.
* Sets ANDROID_SDK_ROOT and ANDROID_HOME inside of the Docker image.
  All the other android related environment varaibles are not needed anymore.
* The user inside docker ('user') now has its home placed at '/home/user'
  instead of '/'.
  This solves many issues:
  * Now $HOME is writable and all its content can be read.
  * As a result environment variables or paths that are derived from $HOME
    work automatically.
  * GRADLE_USER_HOME does not need to be set, instead the external gradle
    cache is mounted to /home/user/.gradle which is picked-up by gradle automatically.
    Note: Since the gradle cache is not relocatable (it hard-codes paths)
          the existing /var/local/container_shared/gradle cannot be used.
          Instead /var/local/container_shared/gradle_cache is used.
  * No need to mount .android, emulator_console_auth_token, and analytics.settings.
    They are created automatically.
  * No need to set JAVA_TOOL_OPTIONS = "-Duser.home=/tmp/" as $HOME/.java
    is now writable.
* The setup stage is not necessary since everything should be provided
  in the Docker container.
* No need to call clearAvdStore manually since ANDROID_AVD_HOME points to
  a directory inside of the containers.
  Thus when the container is stopped everything is cleared automatically.
* Allow to specify id of the kvm group as build paramter.
  That way you can build the docker image locally
* Since the jenkins-android-helper scripts are not used to manage the emulator
  the references to ANDROID_EMULATOR_HELPER_DEBUG and PYTHONUNBUFFERED
  are not needed anymore.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,43 +11,15 @@ pipeline {
 			// corresponding user can be added. This is needed to provide the jenkins user inside
 			// the container for the ssh-agent to work.
 			// Another way would be to simply map the passwd file, but would spoil additional information
-			additionalBuildArgs '--build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g)'
-			// Currently there are two different NDK behaviors in place, one to keep NDK r16b, which
-			// was needed because of the removal of armeabi and MIPS support and one to always use the
-			// latest NDK, which is the suggestion from the NDK documentations.
-			// Therefore two different SDK locations on the host are currently in place:
-			// NDK r16b  : /var/local/container_shared/android-sdk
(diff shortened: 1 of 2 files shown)


ThomasSchwengler:
@amix73 @SebastianGrief @chbinder 
Please have a look at this, I can only merge it after one person other than me approves it as well.

PR #578, opened 2018-11-05
PAINTROID-28, JENKINS-237-track-statistics 

eitzo:
Implements dexcount and apksize tracking and displays plots on jenkins

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -125,6 +125,9 @@ pipeline {
 
 					sh './gradlew stopEmulator clearAvdStore'
 					archiveArtifacts 'logcat.txt'
+
+					plot csvFileName: 'dexcount.csv', csvSeries: [[displayTableFlag: false, exclusionValues: '', file: 'Paintroid/build/outputs/dexcount/*.csv', inclusionFlag: 'OFF', url: '']], group: 'APK Stats', numBuilds: '180', style: 'line', title: 'dexcount'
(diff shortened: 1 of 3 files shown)


mfuchs:
Please adapt your commits so that there are no whitespace changes.
Yes I am not a fan of the tab-usage myself.
Yet tabs were used when creating these file in the first place.

If your changes are intermixed with the whitespace changes then it is very hard to see what the actual new code is.

Jenkinsfile
```diff
@@ -1,150 +1,175 @@
...
+                    junit '**/*TEST*.xml'
+                    step([$class: 'CoberturaPublisher', autoUpdateHealth: false, autoUpdateStability: false, coberturaReportFile: "$JAVA_SRC/coverage*.xml", failUnhealthy: false, failUnstable: false, maxNumberOfBuilds: 0, onlyStable: false, sourceEncoding: 'ASCII', zoomCoverageChart: false, failNoReports: false])
+                    
+                    sh './gradlew stopEmulator clearAvdStore'
+                    archiveArtifacts 'logcat.txt'  
+                    
+                    plot csvFileName: 'dexcount.csv',
+                        csvSeries: [[displayTableFlag: false,
+                        exclusionValues: '',
+                        file: 'Paintroid/build/outputs/dexcount/*.csv',
```
mfuchs: Just in-case.
The build-directory does not survive a ./gradlew clean.
Furthermore the build-directory is not kept for consecutive builds.
Not sure if this is needed here to have previous results.

eitzo: The build-directory does not have to be persistent to plot the previous results.
Should we not move to spaces instead of tabs ? 
We did the same on catroid with the hardwaretests

mfuchs:
Looks good to me.

PR #580, opened 2018-11-08
PAINTROID-2: Fix crash in layer menu

ThomasSchwengler:
* Handle click and longclick events in the listview instead of the item views
* Disable `SplitMotionEvents` in listview
* Adapt tests
* Invalidate listview on touchup properly


ThomasSchwengler:
@chbinder @SebastianGrief Please have a look at this pr

PR #584, opened 2018-11-23
PAINTROID-1_PAINTROID-14 Introduced Headless Retain Fragment instead …

DjordjeR:
…of Paintroid Application Singelton

### DIFF
--- a/Paintroid/src/main/java/org/catrobat/paintroid/PaintroidApplicationFragment.java
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/PaintroidApplicationFragment.java
@@ -1,0 +1,111 @@
+/*
+ * Paintroid: An image manipulation application for Android.
+ * Copyright (C) 2010-2015 The Catrobat Team
+ * (<http://developer.catrobat.org/credits>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


Paintroid/src/main/java/org/catrobat/paintroid/presenter/MainActivityPresenter.java
```diff
@@ -112,9 +112,9 @@
 	private boolean focusAfterRecreate = true;
 
 	public MainActivityPresenter(MainView view, Model model, Navigator navigator, Interactor interactor,
-			TopBarViewHolder topBarViewHolder, BottomBarViewHolder bottomBarViewHolder,
-			DrawerLayoutViewHolder drawerLayoutViewHolder,
-			NavigationDrawerViewHolder navigationDrawerViewHolder, CommandManager commandManager) {
+								TopBarViewHolder topBarViewHolder, BottomBarViewHolder bottomBarViewHolder,
```
ThomasSchwengler: Please revert unrelated style fixes to `MainActivityPresenter`

DjordjeR: I have reverted changes, for some reason Android Studio did this. Should I rename branch to only PAINTROID-14 and make a new request? I tought this solved the problem bacuse I tried going to the Import image from gallery ->  Hamburger menu -> Pocket paint and it no longer chrashes even though it did for play store version. Maybe this is another bug.

ThomasSchwengler: I will have another look at it, maybe this does fix our crashes already.

Paintroid/src/main/java/org/catrobat/paintroid/PaintroidApplicationFragment.java
```diff
@@ -0,0 +1,92 @@
+package org.catrobat.paintroid;
```
ThomasSchwengler: Please add our copyright header

PR #587, opened 2018-12-02
JENKINS-266 Release pipeline (APK)

rainer-lankmayr:
- Added signedRelease buildtype and signing credentials via gradle parameters.

- Added Fastfile and install Fastlane via Docker for upload to Google Play Store.

- Added Jenkinsfiles for release via Jenkins Jobs.

### DIFF
--- a/Jenkinsfile.releaseAPK
+++ b/Jenkinsfile.releaseAPK
@@ -1,0 +1,105 @@
+#!groovy
+
+pipeline {
+    agent {
+        dockerfile {
+            filename 'Dockerfile.jenkins'
+            // 'docker build' would normally copy the whole build-dir to the container, changing the
+            // docker build directory avoids that overhead
+            dir 'docker'
+            // Pass the uid and the gid of the current user (jenkins-user) to the Dockerfile, so a
+            // corresponding user can be added. This is needed to provide the jenkins user inside
+            // the container for the ssh-agent to work.
+            // Another way would be to simply map the passwd file, but would spoil additional information
+            // Also hand in the group id of kvm to allow using /dev/kvm.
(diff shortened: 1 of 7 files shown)


Jenkinsfile.releaseAPK
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    post {
+        always {
+            // clean workspace
+            deleteDir()
+        }
+    }
+}
```
mfuchs: Please add a newline at the end.

app/build.gradle
```diff
@@ -29,12 +33,33 @@ android {
...
+            storeFile file(project["signingKeystore"])
+            storePassword ""
+            keyAlias ""
+            keyPassword ""
+        }
+    }
+
+    if (project.hasProperty("signingKeystore") && project.hasProperty("signingKeystorePassword") &&
+            project.hasProperty("signingKeyAlias") && project.hasProperty("signingKeyPassword")) {
+        android.signingConfigs.signedRelease.storePassword = project.getProperties().get("signingKeystorePassword")
```
mfuchs: I think that using project.property("....") is shorter here.

app/build.gradle
```diff
@@ -29,12 +33,33 @@ android {
         versionCode rootProject.ext.androidVersionCode
         versionName rootProject.ext.androidVersionName
     }
+
+    signingConfigs {
+        signedRelease {
+            storeFile file(project["signingKeystore"])
```
mfuchs: project.property("signingKeystore") would be more explicit

ThomasSchwengler:
test this please

app/build.gradle
mfuchs: What I meant was that property is used after the if. Instead of getProperties().
The hasProperty for the if condition is still necessary as the property call expects the property to exist. See also the failed build.
Sorry for the inconvenience.

PR #594, opened 2018-12-20
(WIP) PAINTROID-15: Refactor Tools Modification of the bottom bar in tools

stojca:


stojca:
I am not sure whether my approach is good, but I removed  Context usage and BottomBar state from BaseTool. Direct access to the DrawingSurface in the BaseTool will be removed with the wrapper class from Paintroid 12.

ThomasSchwengler:
Closed as inactive, ticket moved back to RFD

PR #595, opened 2018-12-20
JENKINS-273 Avoid Jenkins errors on emulator crashes.

mfuchs:
Issue
=====
When the emulator crashes the cover2cover.py script fails.
This leads to a corrupt coverage.xml file and skipping all following steps
in the current stage.
So following build steps are not executed, even if they might succeed.
This is inconsistent: Gradle itself is configured to ignore failed tests,
it also ignores emulator crashes.

Also at the end of the build the cobertura step in the post-block of the
Jenkinsfile fails. Any succeeding steps are skipped here as well.

Solution
========
* More handling is moved to cover2cover.py not only simplifying the Jenkinsile
  but also ensuring that the resulting coverage.xml file is valid.
* cover2cover.py always has an exit code of 0, even if the input
  xml file was corrupt.
* The log parser rules are adapted to detect failures of the cover2cover script.


Jenkinsfile
```diff
@@ -71,14 +71,14 @@ pipeline {
 				// Convert the JaCoCo coverate to the Cobertura XML file format.
 				// This is done since the Jenkins JaCoCo plugin does not work well.
 				// See also JENKINS-212 on jira.catrob.at
-				sh "if [ -f '$JACOCO_UNIT_XML' ]; then ./buildScripts/cover2cover.py $JACOCO_UNIT_XML > $JAVA_SRC/coverage1.xml; fi"
+				sh "./buildScripts/cover2cover.py $JACOCO_UNIT_XML $JAVA_SRC/coverage1.xml"
```
ThomasSchwengler: Should we guard those variables for whitespaces?
```
sh "./buildScripts/cover2cover.py \"$JACOCO_UNIT_XML\" \"$JAVA_SRC/coverage\"
```

ThomasSchwengler:
otherwise LGTM :+1:

Jenkinsfile
mfuchs: I don't think that this is necessary as white spaces in directories is not best-practice.
It is very unlikely that we will encounter white spaces, especially on Jenkins.

PR #602, opened 2019-01-17
PAINTROID-48: Fix crash in onRequestPermissionsResult on API 23

SebastianGrief:


ThomasSchwengler:
Can you please provide some reasoning why this solves the problem?

Is this only affecting API 23?

Adding a delay to fix a crash sounds like it will be flaky, some devices may need more time.
I would rather have the delayed call directly in the `MainActivity`, which would be less code, as you only need it once for any permission to come.

SebastianGrief:
> Can you please provide some reasoning why this solves the problem?

There is an issue for this bug [1] but it's for some reason "Intended behavior" 
An yes it's only on M (as far as I could test/search on the internet).

[1] https://issuetracker.google.com/issues/37067655

SebastianGrief:
> * There is still a fixed delay, this will make it worse in two cases:
>   
>   * For API == 23 it could be too little and we will still get the crash, just not that often, making it hard to debug
>   * For API >23 it will delay the result for no apparent reason.
> 

Only sdk 23 is handled differently now. 
I create a Runnable which waits for the onResume.

ThomasSchwengler:
LGTM :+1:

ThomasSchwengler:
Merged into `release-2.2.0`

PR #603, opened 2019-01-28
PAINTROID-50: Fix crash related to camera permission

ThomasSchwengler:
* Simplify permission logic, don't move uri through control flow if unecessary
* Remove "back to pocket code" request codes, distinguish with activity model as it exposes logic decisions in request codes
* Create base fragment for all dialogfragments needing a presenter callback
* Rewrite PermissionInfoDialog
  * Fix requestCode not being saved
  * Simplify arguments bundle
  * Remove unnecessary calls to `dialog.cancel`
* Fix a bug leading to image not being saved after clicking new image
* Add tests to cover all cases


ThomasSchwengler:
@SebastianGrief, @chbinder please review

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/CameraIntegrationTest.java
```diff
@@ -83,7 +83,8 @@
 	@ClassRule
 	public static GrantPermissionRule grantPermissionRule = GrantPermissionRule.grant(
 			Manifest.permission.WRITE_EXTERNAL_STORAGE,
-			Manifest.permission.READ_EXTERNAL_STORAGE);
+			Manifest.permission.READ_EXTERNAL_STORAGE,
```
thmq: I cannot comment on the line directly because it is not part of the set of changes, but is there a reason why you catch the exception that can occur while creating the picture file (see code snippet below). because what I have found is that just having the `setUp()` method throw the exception gives you the stacktrace plus the underlying reason why this failed opposed to just the message you put into the call to `fail(..)`. 
```
catch(IOException e) {
    fail("Picture file could not be created.");
}
```

ThomasSchwengler: You are right, that is bs, there is no reason for this try-catch to exist.

Paintroid/src/main/java/org/catrobat/paintroid/MainActivity.java
```diff
@@ -436,15 +436,20 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
 	@Override
-	public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
-		presenter.handlePermissionRequestResults(requestCode, permissions, grantResults);
+	public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
+		presenter.handleRequestPermissionResult(requestCode, permissions, grantResults);
 	}
 
 	@Override
-	public void forwardActivityResult(int requestCode, int resultCode, Intent data) {
+	public void baseActivityResult(int requestCode, int resultCode, Intent data) {
```
thmq: since this is called from the presenter whenever you want to have the super class handle the result, what do you think about calling this `superOnActivityResult(..)` or something similar that contains the term `super`?

Paintroid/src/main/java/org/catrobat/paintroid/MainActivity.java
```diff
@@ -436,15 +436,20 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
 		super.onActivityResult(requestCode, resultCode, data);
 	}
 
+	@Override
+	public void baseHandlePermissionRequestResults(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
```
thmq: analogue to  `baseActivityResult(..)` you could consider using the term `super` instead of `base` here to clarify what it does.

Paintroid/src/main/java/org/catrobat/paintroid/MainActivity.java
ThomasSchwengler: Thanks, changed to `super.*`

Paintroid/src/main/java/org/catrobat/paintroid/MainActivity.java
ThomasSchwengler: Thanks, fixed.

Paintroid/src/main/java/org/catrobat/paintroid/presenter/MainActivityPresenter.java
```diff
@@ -664,14 +695,15 @@ public void onSaveImagePostExecute(@SaveImageRequestCode int requestCode, Uri ur
 			case SAVE_IMAGE_DEFAULT:
 				break;
 			case SAVE_IMAGE_FINISH:
-				navigator.finishActivity();
+				if (model.isOpenedFromCatroid()) {
+					navigator.returnToPocketCode(uri.getPath());
```
thmq: We should consider using the uri here instead of a string path. I will create a ticket for this, so let us leave it for now though.

ThomasSchwengler: :+1:

PR #605, opened 2019-01-29
PAINTROID-40: Improve shape tool paste feedback

SebastianGrief:
- changed the highlight color to complementary color of the accent color used

- made border of tool get a little smaller when pressed


ThomasSchwengler:
In my opinion, the `CLICK_TIMEOUT_MILLIS` needs to be `>= 250`, otherwise the feedback is only a little longer than a frame and too little for someone to notice.

Pasting something with the stamp of import png tool don't seem to be using this, should this also be affected?

https://streamable.com/zno78

@VesnaK Is this how you had this in mind?

SebastianGrief:
> In my opinion, the `CLICK_TIMEOUT_MILLIS` needs to be `>= 250`, otherwise the feedback is only a little longer than a frame and too little for someone to notice.
> 
Changed to 250.

> Pasting something with the stamp of import png tool don't seem to be using this, should this also be affected?
> 

Your right, it should... I think the paste command is blocking that... 
Will have a look at it.

ThomasSchwengler:
@chbinder please review as well

PR #606, opened 2019-01-30
PAINTROID-49: Fix crash from callback while in background

ThomasSchwengler:
* Dismiss `IndeterminateProgressDialog` with `dismissAllowingStateLoss`
* Add a `isBusy` flag to `CommandManager`, dismiss in `onCreate` if state was lost (e.g. activity was in background)
* Reject commands while the async command manager is already busy (This fixes a different bug, that appeared when pressing e.g. rotate and mirror buttons at the same time)


ThomasSchwengler:
@SebastianGrief @chbinder @stojca @DjordjeR 
Someone please review this ticket. It's the last release blocking issue.

PR #613, opened 2019-02-08
PAINTROID-39: Fix menus slide buggy

ThomasSchwengler:
* Ignore clicks starting in range of drawer edges
  * This uses the same border width constants as the support library
    navigation drawer
* Add unit tests verifying this behaviour
* Adapt espresso tests


ThomasSchwengler:
@chbinder @SebastianGrief please review

PR #621, opened 2019-02-28
PAINTROID-32: Rename new image

ThomasSchwengler:
* Add an additional 'Discard image' button to the navigation drawer
  * This removes all layers and adds a new one after that
  * As it can be undone, it does not ask the user to save
  * The `savedImageUri` does not change
  * This option is removed for the standalone version, while the integrated version removes the `New image` option.
* Add a `ResetCommand` that removes all layers
* Add local unit tests for the presenter, commandfactory and command
* Move catroid related integration tests to a `catroid` package


ThomasSchwengler:
image

mrmattmueller:
@ThomasSchwengler is there also somewhere an apk for the integrated version? Otherwise I can't test these changes ;)

PR #627, opened 2019-03-04
JENKINS-269: Use the warnings-ng plugin.

mfuchs:
The other static analysis plugins are deprecated and succeeded by warnings-ng.
With JENKINS-55436 fixed we can use warnings-ng ourselves, without breaking
existing jobs.


ThomasSchwengler:
Is this where this new fancy report is coming from?

PR #636, opened 2019-03-15
(WIP) PAINTROID-15: Refactor tools

ThomasSchwengler:
* Refactor tools, remove all view and context dependencies
  * Add a new `ContextCallback` facade class, providing all context dependencies for the tools
  * Add interfaces for tool options, rename and remove any `View` postfixes
  * Move communication with between tools and tool options into callbacks
  * Remove bottom sheet functions from `Tool` interface
  * Refactor `BrushTool` to implement `Tool` directly
  * Add a `ToolOptionsController` and move all common tool bottom sheet functionality there
  * Add a `ToolReference` class, holding the reference to the currently used tool
  * Rename `GeometricFillTool` to `ShapeTool`
  * Rename `DrawTool` to `BrushTool`
* Rewrite all instrumented tool junit tests to remove activity dependencies
  * Split up some tests into `.*Command` and `.*Algorithm` tests
* Remove obsolete tests

PR #637, opened 2019-03-20
PAINTROID-1: Fix crash when second MainActivity in same process

ThomasSchwengler:
* Add a `ToolReference` class, replacing the `public static Tool currentTool`
* Rename `DrawTool` to `BrushTool`
* Rename `GeometricFillTool` to `ShapeTool`
* Rename `.*Tests` classes to `.*Test`
* Fix all failing tests and utility test classes


ThomasSchwengler:
This is now a precondition to [PAINTROID-15](https://jira.catrob.at/browse/PAINTROID-15) (#636).

PR #639, opened 2019-03-22
PAINTROID-15: Refactor tools

ThomasSchwengler:
* Refactor tools, remove all context dependencies and tool option states
  * Add a new `ContextCallback` facade class, providing all context dependencies for the tools
  * Move communication with between tools and tool options into callbacks
  * Remove bottom sheet functions from `Tool` interface
  * Add a `ToolOptionsController` and move all common tool bottom sheet functionality there
  * Move common tool constants to a `Constants` class


ThomasSchwengler:
@Hiebl I added `enable()` and `disable()` functions to the `ToolOptionsController`. These were called `toggleToolOptions` before which made no sense at all.
This disables all tool options while in fullscreen and fixes the testcases we saw failing while reviewing this afternoon.

PR #656, opened 2019-04-17
 PAINTROID-72 Preview Line no longer starts at previous line endpoint 

StanglTobias:


Paintroid/src/main/java/org/catrobat/paintroid/tools/implementation/DefaultWorkspace.java
```diff
@@ -45,6 +48,18 @@ public boolean contains(PointF point) {
 		return point.x < getWidth() && point.x >= 0 && point.y < getHeight() && point.y >= 0;
 	}
 
+	@Override
+	public boolean intersectsWith(RectF rectangle) {
+		Canvas canvas = new Canvas();
```
ThomasSchwengler: This is unecessary.
You only need the width and the height of the workspace with `getWidth()` and `getHeight()` respectively.
In this case you don't even need a rectangle (and not the huge amount of memory)

PR #664, opened 2019-04-29
PAINTROID-67: Hand Tool

chbinder:


Paintroid/src/main/res/values/string.xml
```diff
@@ -62,6 +63,7 @@
     <string name="help_content_shape">Choose a shape and tap it to insert the selected shape</string>
     <string name="help_content_layer">Create new layers or modify existing ones</string>
     <string name="help_content_color_chooser">Select or adjust a color</string>
+    <string name="help_content_hand">PLACEHOLDER</string>
```
ThomasSchwengler: Please add some descriptive text, this will show up in the tutorial/intro

PR #687, opened 2019-06-11
PAINTROID-93 Move sidebar into more options

StanglTobias:


mrmattmueller:
The sidebar is still there in landscape mode (i.e., if you are in landscape and swipe in from the left side). As stated in the acceptance criteria it should be totally removed.

PR #695, opened 2019-07-02
PAINTROID-94 Back to PocketCode Arrow

StanglTobias:
* [x] Depends on #704


ThomasSchwengler:
Please resolve conflicts

wslany:
I just tried it, but it looks as follows:
image

When I go into landscape mode, the back button appears, but 
image

When I turn the phone back into portrait mode, the arrow button remains, but on the wrong side:
image

wslany:
Thanks!

PR #698, opened 2019-07-09
PAINTROID-97: Changing single character in color chooser HEX

SebastianGrief:
- Cursor reset to last position after EditText update


geolas94:
rebuilding on jenkins passes all tests

PR #711, opened 2019-10-02
add packages to dockerfile

StanglTobias:


ThomasSchwengler:
Merged as there is no need for a PO review

PR #712, opened 2019-10-07
PAINTROID-111: Rate us button

SebastianGrief:
- Added rate us button which starts new PlayStore activity or if the PlayStore isn't installed the browser


Paintroid/src/main/java/org/catrobat/paintroid/presenter/MainActivityPresenter.java
```diff
@@ -659,4 +659,9 @@ public void actionCurrentToolClicked() {
 			}
 		}
 	}
+
+	@Override
+	public void rateUsClicked() {
+		navigator.rateUsClicked();
```
ThomasSchwengler: Please add a test for this

ThomasSchwengler:
Please rebase

PR #713, opened 2019-10-07
PAINTROID-112: disable help

SebastianGrief:


Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/MenuFileActivityIntegrationTest.java
```diff
@@ -203,7 +199,7 @@ public void testLoadImageDialogOnBackPressed() {
 		onDrawingSurfaceView()
 				.check(matches(isDisplayed()));
 	}
-
+/* Enable if help button is enabled again
```
ThomasSchwengler: Please remove code instead of commenting out. We can restore it from VCS at a later time if needed.

ThomasSchwengler: Instead, please add a test verifying that the option is gone.

wslany:
Thanks!

PR #728, opened 2020-03-15
PAINTROID-133 Second launcher to simulate Catroid

StanglTobias:
-added second launcher for development build to simulate launch from catroid


wslany:
Thanks!

PR #786, opened 2020-09-29
PAINTROID-163: Save temporary copies automatically

FromG:
Added 2 methods in FileIO:
 One for saving a temporary file into the internal storage.
 One for checking if there is a temporary file.
Added methods in MainActivity:
 Created a thread which checks for idle time.
 If a set amount of time is reached the file will be saved to internal storage.
Added new File
Added a constant.
Added Tests in espresso/catroid folder.
 
**Current Behaviour :** If the user does not do anything for a certain amount of time, it will create a temporary file of the current
picture. When the user idles again the temporary file will be overriden with the current picture. If the user then closes the app without saving the temporary file remains in internal storage. On opening the app once again the user will get a window
which tells him that there is a temporary file available and asks him if he wants to open the file. The file will be opened if the user
accepts. If the user saves the file the temporary file will be deleted.

Tests: 
API:27
Landscapeintegrationtests sometimes fail locally because of the single click error.
Testtooloptionsdiappearwhenbackpressed test fails on develop and this branch.
API 29:
Landscapeintegrationtests sometimes fail locally because of the single click error.
testcommandundoredo and testcheckboxsizeandcontentafterformattingtodubaicandstc tests fail
on develop and this branch.

- [ ] Post a message in the *#paintroid* [Slack channel](https://catrobat.slack.com) and ask for a code reviewer

### DIFF
--- a/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/catroid/TemporaryFileSaveTest.java
+++ b/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/catroid/TemporaryFileSaveTest.java
@@ -1,0 +1,105 @@
+package org.catrobat.paintroid.test.espresso.catroid;
+
+import android.content.Intent;
+import android.graphics.Bitmap;
+import android.graphics.Color;
+
+import org.catrobat.paintroid.MainActivity;
+import org.catrobat.paintroid.R;
+import org.catrobat.paintroid.common.Constants;
+import org.catrobat.paintroid.test.espresso.util.BitmapLocationProvider;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+import androidx.test.espresso.intent.rule.IntentsTestRule;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
+import static org.catrobat.paintroid.test.espresso.util.wrappers.DrawingSurfaceInteraction.onDrawingSurfaceView;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import static androidx.test.espresso.Espresso.onView;
+import static androidx.test.espresso.action.ViewActions.click;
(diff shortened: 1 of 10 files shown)


geolas94 (approved):
lgtm

wslany (changes requested):
I tried to simulate the situation but was unable to recover the data. I draw something, then left the device idle for 5 up to 10 seconds, then force-closed the app, and then restarted it. The old drawing was not recovered. According  to https://github.com/Catrobat/Paintroid/blob/c145a1f710970dfed3d13064174a6b046c0dfbb6/Paintroid/src/main/java/org/catrobat/paintroid/MainActivity.java#L235 the time after which the state is saved is 2 minutes of idle time. This is much too long. Imagine someone is continuously drawing for 30 minutes, then the app is force closed before the 2 minutes of idle time. Everything will be lost... 

Also, once I tried the stuff with the 2 minutes of wait, then the force close, then accpeting the recovered date after the restart of the app, then continuing to work again on the image for 3 minutes, and then force closing the app again. After starting it once more, I again was offered to recover the image, but the recovered state was of course the first one, so the last 3 minutes of drawing where lost.

So, the logic is insufficient to solve the issue and the loos of data of real users. What should actually happen to make sense for the users, is: The temporary saving should happen much more frequently. Whenever the app is idle for 2 seconds, there should be an attempt to save the state. At the longest, the state shall be saved every 60 seconds, even if the device is not idle. And the saving shall happen in a parallel thread. 

Also, only bitmaps are saved, even though the ticket required the image to be stored in the internal data format, i.e., including layers, undo information etc. 

Also note that the chosen approach is highly inefficient, since the whole bitmap has to be stored each time. And it can even fail when the writing of the temporary file is interrupted, so there would have to be two temporary copies, and only once the second one has been written can the first one destroyed and the second one renamed to the first one. Since the saving is not done in a thread, in case it takes a long time, the app can become unresponsive until the data has been completely saved.

But there actually is a safer and much more efficient method that is provided by the Android system, see https://developer.android.com/topic/libraries/architecture/saving-states for details. 

Also, there must be at least two more testcases, as follows:
- [ ] when the temporary data is overwritten because another, newer version exists, the retrieved data shall be the latest, i.e., the new version.
- [ ] when the thread was force-closed (this can be simulated and mocked), and the activity is then resumed, the last saved state shall be recovered.

PR #789, opened 2020-10-14
PAINTROID-148 Export as jpg (old branch)

FromG:
**new branch for ticket : https://github.com/Catrobat/Paintroid/pull/809  !!!!**

created jpg png information dialog and added functioning.

Save image option:
If there is currently no image saved the dialog shows up with the standard values.(filename = "image", format = "jpg")
If there is a saved image the dialog will always show the saved image information.

Save copy option:
The copy option will always show the standard format. ("image5" "jpg");

when opened from catroid(e.g. pocketcode):
the dialog will not show. it will save automatically as png.

Overwriting: 
When the user wants to overwrite an existing file a dialog pops up which will warn the user about it.

Behavior:
Whenever you save an Image a Save dialog shows up where you can choose between
format and if you choose jpg you can then change the quality(also the transparent background will be changed to white when the picture gets saved). You can also tap
on the info button to get information about the file format.


Electronix1337 (changes requested):
When I create a new file and save that as png (filename1.png for example) and then save that file again with jpg, the file name is filename1.jpg which should be the correct behaviour but when i save that file again as .png the filename is filename1.png (1) but the file type is "BIN" and I can't open that file.

Electronix1337 (changes requested):
Mentioned bug is gone, about the layout I would ask @wslany if current layout is fine or if it should be a bit more dense. I would also consider using the incremented default filename when user is loading a file because otherwise the app would generate a "Bin" file. Rest lgtm.

Electronix1337 (approved):
Functionality lgtm about the layout I would let @wslany decide if changes are necessary.

Paintroid/src/main/res/values/string.xml
```diff
@@ -143,11 +145,21 @@
     <string name="dialog_loading_image_failed_text">Not a valid image</string>
     <string name="dialog_settings">Settings</string>
 
+    <string name="dialog_save_image_name">Image name</string>
+    <string name="dialog_save_image_format">Image format</string>
+
+    <string name="dialog_save_jpg_option_quality">Quality:</string>
+    <string name="dialog_save_options_size">Size:</string>
```
wslany, line 152: Where is this string used?

Paintroid/src/main/res/values/string.xml
```diff
@@ -143,11 +145,21 @@
...
+
+    <string name="pocketpaint_jpg_title_dialog" translatable="false">jpg</string>
+    <string name="pocketpaint_jpg_message_dialog">Takes up <b>minimal storage space</b>. <b>No Transparency</b> is remembered</string>
```
wslany, line 155:
```suggestion
    <string name="pocketpaint_jpg_message_dialog">Takes up <b>minimal storage space</b>. <b>No transparency</b> is remembered.</string>
```

Paintroid/src/main/res/values/string.xml
```diff
@@ -143,11 +145,21 @@
...
+    <string name="pocketpaint_png_title_dialog" translatable="false">png</string>
+    <string name="pocketpaint_png_message_dialog">Lossless compression. Saves transparency</string>
```
wslany, line 157:
```suggestion
    <string name="pocketpaint_png_message_dialog">Lossless compression. Transparency is preserved.</string>
```

wslany (changes requested):
Thanks!

Please make sure that the "Rate Pocket Paint" dialog appears **_after_** the saving interaction is complete, not before.

image

Also, how the buttons at the bottom are placed in the dialog seems not to be material design conform AFAIK. Please consult with @ThorstenBandel about the correct placement.

image

Also, transparent areas in the original image shall be rendered in white in jpg files, not in black as it is currently done. 
image

org.catrobat.paintroid.test.espresso.catroid.OpenedFromPocketCodeNewImageTest.testSave (from org.catrobat.paintroid.test.espresso.BitmapIntegrationTest) is failing every run because of 
```
Stacktrace
org.junit.ComparisonFailure: expected:</storage/emulated/0/[Android/data/org.catrobat.paintroid.test/files/]Pictures/Look123.png> but was:</storage/emulated/0/[]Pictures/Look123.png>
at org.junit.Assert.assertEquals(Assert.java:115)
at org.junit.Assert.assertEquals(Assert.java:144)
at org.catrobat.paintroid.test.espresso.catroid.OpenedFromPocketCodeNewImageTest.testSave(OpenedFromPocketCodeNewImageTest.java:122)
```

See some of the inlined additional comments in the code.

Paintroid/src/main/res/values/string.xml
FromG: in dialog_pocketpaint_save.xml and dialog_pocketpaint_save_jpg.sub_dialog.xml .

Electronix1337 (approved):
lgtm, layout looks much better, dialog is displayed after save dialog and jpg images have white background.

Electronix1337 (changes requested):
testBrushToolBackPressedFromCatroidAndUsePicture in ToolOnBackPressedIntegrationTest.java can be deleted in my opinion because it tests the same as OpenedFromPocketCode and testBrushToolBackPressedFromCatroidAndDiscardPicture should be deleted because this is not possible anymore due to implicit saving. I would also suggest to adjust the OpenedFromPocketCodeNewImageTest.testSave because the save dialog should not show up and therefore the test has to be rewritten. In the current version the "openedfromcatroid" flag is not even set. And we should clarify if implicit saving from catroid should be done with png/jpg or if we let the user decide which would make the implicit saving kind of redundant? @wslany

wslany (changes requested):
This I like very much!! => 
image

**But:** When using the paintroid lib in Catroid => https://jenkins.catrob.at/job/Paintroid/job/PR-789/40/artifact/catroid/build/outputs/apk/catroid/debug/catroid-catroid-debug.apk and saving an edited (or newly created) image, after tapping on save, another 2nd save dialog with jpg as default is shown. _This is incorrect._ The correct behavior shall be: when returning from Paintroid to Catroid, only the png image should automatically be handed back, and there shall be no dialog. Here how it looks like at the moment after already having tapped on SAVE once: 
image

In Pocket Paint there are two options, Save image and Save copy, but they seem to do exactly the same. Instead, I think it would be better for the "Save image" option to show the original filename of the image in the dialog, with the original filetype of the image preselected, and then, if the file name is not changed by the user,  the file shall be written *over* the file that was the original. This seems to make more sense IMHO.
image

The following is not due to this PR, but is related, and therefore noted here: After using the "export" option in Paintroid lib from within Catroid, there is a short toast message showing "Copy":
image
I am not sure where this text comes from, especially since there is no "Copy" text in string.xml of Paintroid, and this part clearly is still executed in the Paintroid library part. Anyway, this text is misleading since the menu option is "Export", and nothing is said about copying. Note that the text is different in the standalone version, Pocket Paint. But there is already a ticket in development through which hopefully these toast messages shall anyway be improved for all cases, namely https://jira.catrob.at/browse/PAINTROID-141 @geolas94 please correct me if I am wrong, so I guess nothing has to be done here. (@geolas94 , please add your full name to your account).

geolas94 (approved):
lgtm

Paintroid/src/main/java/org/catrobat/paintroid/UserPreferences.java
```diff
@@ -15,6 +16,17 @@ public boolean getBoolean(String tag, boolean defaultReturnValue) {
...
+	public int getInt() {
+		return preferences.getInt(IMAGE_NUMBER_SHARED_PREFERENCES_TAG, 0);
+	}
+
+	public void setInt(String tag, int value) {
+		preferences
+			.edit()
+			.putInt(IMAGE_NUMBER_SHARED_PREFERENCES_TAG, value)
+			.apply();
+	}
```
StanglTobias, line 28: Please change to getter and setter for the corresponding variable not just Int.

Also do the same for getBoolean/setBoolean.

StanglTobias (changes requested):

Paintroid/src/main/java/org/catrobat/paintroid/dialog/SaveInformationDialog.java
```diff
@@ -0,0 +1,203 @@
...
+				dismiss();
+			}
+		});
+
+		cancelButton.setOnClickListener(new View.OnClickListener() {
+			@Override
+			public void onClick(View v) {
+				dismiss();
+			}
+		});
```
StanglTobias, line 152: Use setPositiveButton and setNegativeButton in with AlertDialog.Builder instead.

StanglTobias (changes requested):
app crashes when save dialog is open and the device is rotated.
When jpg is selected. The quality test should be aligned vertically to the slider.

StanglTobias:
To be discussed @wslany  @ThorstenBandel:
Should overriding an image without notification be available if the uses chooses an existing image name?

wslany:
> To be discussed @wslany @ThorstenBandel:
> Should overriding an image without notification be allowed if the uses chooses an existing image name?

I think there should be a dialog in that case, maybe with a short hint that saving a "copy" is also available as an option in the overflow menu. Or maybe even as a choice in the dialog.

Electronix1337 (approved):
lgtm, I think it's efficient enough to give a hint that "save copy" is the alternative if the user does not want to overwrite the file.

wslany (changes requested):
Not sure why, but the pull-down with the choice between jpg and png seems empty:
image
Could that be to the conflict resolution I commited here: 
image

Also, when called from Catroid, the "Export" menu option is incorrectly _not_ asking for jpg versus png. The dialog _should_ happen when the user chooses "Export". It should only _not_ be a choice when returning from Paintroid to Catroid, since there it should always be png, but that's different from the "Export" option. 

Sorry, there now are also some static analysis errors:
https://jenkins.catrob.at/job/Paintroid/job/PR-789/68/analysis/

PR #792, opened 2020-10-23
PAINTROID-151: Redesign copy&paste tool options

StanglTobias:


Electronix1337 (changes requested):
I really like the new design of the stamp tool options! But some (hopefully) minor things have to be changed in my opinion. By the way the deletion of the long click is reasonable in my opinion. 

1) If we copy a box content and rotate the phone the paste button is disabled, even though the stamp tool box has the content before rotating.
2) At least for me on API <= 23 the new buttons are not displaying any text at all. [https://ibb.co/Z1Hc5nw](url) and I would also suggest to have a small space between the "Save" and "Discard" button.
3) I'm not quite sure where this happens, but the selected color in the color picker option is not displayed in the background of the "Apply" button. Background of "Apply" button always stays in the same color except when choosing "transparent" then the color is changed to white otherwise, only the text is changed from black to white depending on the selected color.
4) Is there a special reason why "Cut" option was not implemented? We could create an own ticket for that if it's too much for one ticket but I think the feature can be extremely helpful for users.
5) In my opinion it's quite unintuitive how the tool options overall behave, why not just close the tool options window if user presses back and they are currently opened? And we might consider disabling the onclick for stamp tool options, because the options have to be present all the time in order to actively use the tool, so I don't know if hiding the stamp tool options makes sense. It's definitely just a matter of taste but for me it's rather disturbing than helpful.

StanglTobias:
>     4. Is there a special reason why "Cut" option was not implemented? We could create an own ticket for that if it's too much for >         one ticket but I think the feature can be extremely helpful for users.

There is a seperate ticket for the cut option.


>     5. In my opinion it's quite unintuitive how the tool options overall behave, why not just close the tool options window if user 
>         presses back and they are currently opened? And we might consider disabling the onclick for stamp tool options, because >         the options have to be present all the time in order to actively use the tool, so I don't know if hiding the stamp tool options 
>         makes sense. It's definitely just a matter of taste but for me it's rather disturbing than helpful.

I totally agree with this, but this has nothing in common with this PR. It should be treated in a separate ticket since this one PR did not change the behavior to this.

Paintroid/src/main/java/org/catrobat/paintroid/presenter/MainActivityPresenter.java
```diff
@@ -431,8 +431,6 @@ public void onBackPressed() {
 			drawerLayoutViewHolder.closeDrawer(Gravity.END, true);
 		} else if (model.isFullscreen()) {
 			exitFullscreenClicked();
-		} else if (toolController.toolOptionsViewVisible()) {
-			toolController.hideToolOptionsView();
```
Electronix1337, line 435: Errors are all fixed, but back to 5) aren't that the lines of code for instantly setting the brush tool when pressing back? On develop branch the behavior is as desired: If option window is open and user presses back, the option window is closed. I would also prefer that brush tool option window gets closed when user touches the drawing surface but that's just a matter of taste again. Going to put it into PO review and let Wolfgang decide if changes are needed or not.

Electronix1337 (approved):

Electronix1337 (changes requested):
Actually one little thing has to be fixed, for some reason the "transparent" color is displayed as white now. Maybe fix this before I drag the ticket into PO review.

Colorpicker on develop branch:

color_picker_normal

Colorpicker in this ticket: 

color_picker_transparent_wrong

StanglTobias:
all tests pass and my local device.

Electronix1337 (approved):
lgtm, flaky tests on Jenkins, seem to depend on the slave/state of Jenkins, whether these tests fail or not.

wslany:
Thanks!

PR #797, opened 2020-11-09
PAINTROID-196 Share images from other apps

Electronix1337:
Implemented sharing/editing pictures with Paintroid and the app shows up when the user clicks on any image and tries to share/edit it. A bitmap with the picture will be created (equal to load image) and when user saves the image a new image is saved. When the user clicks on the share/edit button the MainActivity gets invoked and identifies, that someone tries to share data with us. So each time there will be a new instance of the app.
https://jira.catrob.at/browse/PAINTROID-196

### DIFF
--- a/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/FileFromOtherSourceIntegrationTest.java
+++ b/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/FileFromOtherSourceIntegrationTest.java
@@ -1,0 +1,141 @@
+package org.catrobat.paintroid.test.espresso;
+
+import android.content.ContentResolver;
+import android.content.ContentValues;
+import android.content.Intent;
+import android.graphics.Bitmap;
+import android.net.Uri;
+import android.os.Build;
+import android.os.Bundle;
+import android.os.Environment;
+import android.provider.MediaStore;
+import android.util.Log;
+
+import org.catrobat.paintroid.FileIO;
+import org.catrobat.paintroid.MainActivity;
+import org.catrobat.paintroid.test.espresso.util.EspressoUtils;
+import org.catrobat.paintroid.tools.ToolType;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Objects;
+
+import androidx.test.rule.ActivityTestRule;
+import androidx.test.rule.GrantPermissionRule;
+
+import static org.catrobat.paintroid.test.espresso.util.wrappers.ToolBarViewInteraction.onToolBarView;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class FileFromOtherSourceIntegrationTest {
+
+	private static ArrayList<File> deletionFileList = null;
+
+	@Rule
+	public ActivityTestRule<MainActivity> launchActivityRule = new ActivityTestRule<>(MainActivity.class);
+
+	@ClassRule
+	public static GrantPermissionRule grantPermissionRule = EspressoUtils.grantPermissionRulesVersionCheck();
+
+	private ContentResolver resolver;
+
+	@Before
+	public void setUp() {
+		onToolBarView().performSelectTool(ToolType.BRUSH);
+		deletionFileList = new ArrayList<>();
+		resolver = launchActivityRule.getActivity().getContentResolver();
+	}
+
+	@Test
+	public void testGetSharedPictureFromOtherApp() {
+		Intent intent = new Intent();
+		Uri receivedUri = createTestImageFile();
+		Bitmap receivedBitmap = null;
+
+		try {
+			receivedBitmap = FileIO.getBitmapFromUri(resolver, receivedUri);
+		} catch (Exception e) {
+			Log.e("Can't read", "Can't get Bitmap from File");
+		}
+
+		Objects.requireNonNull(receivedBitmap);
+		intent.setData(receivedUri);
+		intent.setType("image/png");
+		intent.setAction(Intent.ACTION_SEND);
+		intent.putExtra(Intent.EXTRA_STREAM, receivedUri);
+
+		launchActivityRule.launchActivity(intent);
+		Intent mainActivityIntent = launchActivityRule.getActivity().getIntent();
+
+		String intentAction = intent.getAction();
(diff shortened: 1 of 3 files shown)


rancidfrog:
You should also add **edit** ability.
I.E.
```
<intent-filter>
<action android:name="android.intent.action.EDIT" />
<category android:name="android.intent.category.DEFAULT" />
<data android:mimeType="image/*" />
</intent-filter>
<intent-filter> <action android:name="action_nextgen_edit" />
<category android:name="android.intent.category.DEFAULT" />
 <data android:mimeType="image/*" />
</intent-filter> 
```

Electronix1337:
> You should also add **edit** ability.
> I.E.
> 
> ```
> <intent-filter>
> <action android:name="android.intent.action.EDIT" />
> <category android:name="android.intent.category.DEFAULT" />
> <data android:mimeType="image/*" />
> </intent-filter>
> <intent-filter> <action android:name="action_nextgen_edit" />
> <category android:name="android.intent.category.DEFAULT" />
>  <data android:mimeType="image/*" />
> </intent-filter> 
> ```

Well this would only be needed if we want to overwrite the shared picture if I'm not mistaken? But as far as i know we want the same behavior as loading a image which doesn't overwrite the original file.

FromG (approved):
lgtm

rancidfrog:
@Electronix1337 
Not necessarily. It is used by gallery apps to edit any photograph or screenshots, saving edits to separate file.

For example:
I take a screenshot and get the options,



I choose edit,


After choosing gallery, any edits I make is saved to new file, not over-writen

So, having an option to use Paintroid to edit woukd be perfect. As I use it to edit screenshots and photographs before sending.

wslany (commented):
There is a kind of strange unexpected "feature", as follows: When only Catroid is installed, and the user wants to share an image, the corresponding Catroid flavor will now also be listed as a possible app to open the image. When one selects it, however, Pocket **Paint** is opened with the image, *even though it is not installed as a separate app!* =>
image
image

@Electronix1337 is this easy to correct? I mean, it is nothing bad, it is just unexpected and kind of miraculous. Could potentially be disturbing for users to suddenly see an unknown app here. Not sure. No big harm done, but if it can be avoided, why not. Anyway, LMK. 

Also, I think @rancidfrog has a very good point --- it makes sense to also be able to edit the original file and give the user the option to either overwrite that original or alternatively to create a new image. I mention this in https://github.com/Catrobat/Paintroid/pull/789#pullrequestreview-535663381 --- we already have "Save image" and "Save copy". Even with "Save image", we can offer the new dialog for saving various formats (jpg, png, internal format) but the filename would be the original one, and also the original path, so the original file can be overwritten if the user does not change the filename or the filetype. For "Save copy", it would instead be a new filename, and the standard path for images. AFAIK we haven't planned to make the path editable, so that could be an issue for when the file is just "edited", so I guess that we need to remember that too, right. I think for now we could nevertheless relegate all of that to a new ticket, I would say, so we could merge this PR as it is. Please correct me if doing it in another way would be better. 

@Electronix1337 can you comment on the above? As I said, I am willing to merge this PR as it is if we better put the mentioned other things into new tickets, or don't bother with them. Just LMK.

Electronix1337:
> There is a kind of strange unexpected "feature", as follows: When only Catroid is installed (https://jenkins.catrob.at/job/Paintroid/job/PR-797/10/artifact/catroid/build/outputs/apk/catroid/debug/catroid-catroid-debug.apk) , and the user wants to share an image, the corresponding Catroid flavor will now also be listed as a possible app to open the image. When one selects it, however, Pocket **Paint** is opened with the image, _even though it is not installed as a separate app!_ =>
> image
> image
> 
> @Electronix1337 is this easy to correct? I mean, it is nothing bad, it is just unexpected and kind of miraculous. Could potentially be disturbing for users to suddenly see an unknown app here. Not sure. No big harm done, but if it can be avoided, why not. Anyway, LMK.
> 
> Also, I think @rancidfrog has a very good point --- it makes sense to also be able to edit the original file and give the user the option to either overwrite that original or alternatively to create a new image. I mention this in [#789 (review)](https://github.com/Catrobat/Paintroid/pull/789#pullrequestreview-535663381) --- we already have "Save image" and "Save copy". Even with "Save image", we can offer the new dialog for saving various formats (jpg, png, internal format) but the filename would be the original one, and also the original path, so the original file can be overwritten if the user does not change the filename or the filetype. For "Save copy", it would instead be a new filename, and the standard path for images. AFAIK we haven't planned to make the path editable, so that could be an issue for when the file is just "edited", so I guess that we need to remember that too, right. I think for now we could nevertheless relegate all of that to a new ticket, I would say, so we could merge this PR as it is. Please correct me if doing it in another way would be better.
> 
> @Electronix1337 can you comment on the above? As I said, I am willing to merge this PR as it is if we better put the mentioned other things into new tickets, or don't bother with them. Just LMK.

I'm going to look at it but i think/hope it should not be too trick to correct this. I'll try and let you know if it's trickier than initially thought we can create an own ticket for that purpose.

Electronix1337:
Implemented that editing images is also possible. Has the same behavior as sharing the file due to the fact that we can't overwrite the existing file because we don't know if we have permissions. Tried overwriting with the google photos app but was not possible. Also might have found a solution for the unexpected "feature" but IMO we have to adjust that in the AndroidManifest.xml file of catroid/src/main/AndroidManifest.xml which i think can only be tested after this PR gets merged into develop and i then have to create a PR in Catroid in order to test if the fix works.

The code for the fix would be something like:
`        <activity android:name="org.catrobat.paintroid.MainActivity" >
             <intent-filter tools:node="remove">
                <action android:name="android.intent.action.SEND" />
                <action android:name="android.intent.action.EDIT" />
                <category android:name="android.intent.category.DEFAULT" />
                <data android:mimeType="image/*" />
            </intent-filter>
        </activity>`
before closing the </application> tag.

rancidfrog:
For me on android 10, save copy and save image do the exact same thing, save a new copy in pocket paint folder.
No overwrite occurs at all.
So, was going to open issue saying having two buttons that do the same thing is useless,

Electronix1337:
> For me on android 10, save copy and save image do the exact same thing, save a new copy in pocket paint folder.
> No overwrite occurs at all.
> So, was going to open issue saying having two buttons that do the same thing is useless,
> 

That's true, it's the same for load image because we cant assure that we have permissions for that Uri. Save copy has only a real usage when creating a new image and after editing the user wants to cache the progress.

wslany:
So yeah, it still shows up, but OK, let's forget it, it is more a (quietly hidden and weird ;-) )feature of a kind of ghost-app appearance of Pocket Paint, than a bug.
image
and (without being installed!):
image

PR #802, opened 2020-12-15
PAINTROID-122 Keep image file size after editing

Electronix1337:
File size of image highly depends on the format of the image. Png images are not really affected but jpg are because our standard compression format is PNG. Will only implement this for Catroid after implicit saving is merged because there users can't decide in which format they want to save their file. For standalone version this is not necessary because user can choose the desired compression format and if they want to use a JPG compression for a PNG file or vice versa, they should be able to in my opinion. As soon as a file is opened from catroid we are checking the mime type and set the correct compression format.

Behavior can be tested as follows: 
1) Download a .jpg picture, the bigger the better
2) Load it into pocket code
3) Make a copy of the picture and rotate the copy. 
4) Now you can see that the size is much lower than before this PR.
5) Reason for that is that we always compress with PNG as compress format, even if the picture is a JPG one.

Files size after rotation was dramatically increased for .jpg pictures before:
image

Files sizes after rotating keep nearly the same after changes:
image

https://jira.catrob.at/browse/PAINTROID-122

### DIFF
--- a/Paintroid/src/main/java/org/catrobat/paintroid/FileIO.java
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/FileIO.java
@@ -105,7 +105,6 @@ public final class FileIO {
 			ContentValues contentValues = new ContentValues();
 			contentValues.put(MediaStore.Images.Media.DISPLAY_NAME, fileName);
 			contentValues.put(MediaStore.Images.Media.MIME_TYPE, "image/*");
-
 			contentValues.put(MediaStore.Images.Media.RELATIVE_PATH, Environment.DIRECTORY_PICTURES);
 
 			imageUri = resolver.insert(MediaStore.Images.Media.EXTERNAL_CONTENT_URI, contentValues);
--- a/Paintroid/src/main/java/org/catrobat/paintroid/ui/MainActivityNavigator.java
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/ui/MainActivityNavigator.java
@@ -23,9 +23,11 @@ import android.annotation.SuppressLint;
 import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.database.Cursor;
 import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.Build;
+import android.provider.OpenableColumns;
 import android.view.Gravity;
 import android.widget.Toast;
 
@@ -58,6 +60,8 @@ import org.catrobat.paintroid.dialog.SaveBeforeNewImageDialog;
 import org.catrobat.paintroid.dialog.SaveInformationDialog;
 import org.catrobat.paintroid.tools.ToolReference;
 import org.catrobat.paintroid.ui.fragments.CatroidMediaGalleryFragment;
+
+import java.util.Objects;
 
 import androidx.appcompat.app.AppCompatDialog;
 import androidx.appcompat.app.AppCompatDialogFragment;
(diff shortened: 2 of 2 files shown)


FromG (approved):

wslany:
I'm merging this because it will do some good, but there still are things left to do:

* Tests are missing
* See the following blowups, from the test images on browserstack:

image

1. The first copies respectively are just turned => this doubles or almost in the second case triples their sizes.
2. The second copies, without turning, have two red lines added in an x pattern in the center => this also more or less doubles their sizes.
3. The third copies are something else: basically any edit is done on the copy of the original, then in Catroid the undo is executed that was implemented in https://jira.catrob.at/browse/CATROID-761 --- wow, but this is actually a problem in Catroid. However, if you have any idea how it could be solved there, I would be happy.  The offending lines of code can be found here:
https://github.com/Catrobat/Catroid/pull/3898/files#diff-a85a0c775283004b0e5d0331265f265168f88711ec6c609f0d2b2c3f9e0985d5

I guess the problem is that the image is kept as a bitmap here:
https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/LookListFragment.java#L69-L70
and https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/LookListFragment.java#L222-L223

The resulting project (as a "debug" upload, i.e., only visible with debug versions) is here: https://share.catrob.at/pocketcode/project/d4e65816-4e88-11eb-a92d-005056a36f47

Here the details on the sample files:
image
image

wslany:
Thanks!!

Electronix1337:
> I'm merging this because it will do some good, but there still are things left to do:
> 
>     * Tests are missing
> 
>     * See the following blowups, from the test images on browserstack:
> 
> 
> image
> 
>     1. The first copies respectively are just turned => this doubles or almost in the second case triples their sizes.
> 
>     2. The second copies, without turning, have two red lines added in an x pattern in the center => this also more or less doubles their sizes.
> 
>     3. The third copies are something else: basically any edit is done on the copy of the original, then in Catroid the undo is executed that was implemented in https://jira.catrob.at/browse/CATROID-761 --- wow, but this is actually a problem in Catroid. However, if you have any idea how it could be solved there, I would be happy.  The offending lines of code can be found here:
>        https://github.com/Catrobat/Catroid/pull/3898/files#diff-a85a0c775283004b0e5d0331265f265168f88711ec6c609f0d2b2c3f9e0985d5
> 
> 
> I guess the problem is that the image is kept as a bitmap here:
> https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/LookListFragment.java#L69-L70
> and https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/LookListFragment.java#L222-L223
> 
> The resulting project (as a "debug" upload, i.e., only visible with debug versions) is here: https://share.catrob.at/pocketcode/project/d4e65816-4e88-11eb-a92d-005056a36f47
> 
> Here the details on the sample files:
> image
> image

For 1. and 2. I don't have an idea right now, why those image sizes are increased so dramatically, if I've got time I'm going to look at it. For 3. the lines https://github.com/Catrobat/Catroid/blob/develop/catroid/src/main/java/org/catrobat/catroid/ui/recyclerview/fragment/LookListFragment.java#L222-L223 are most likely the problem as the image always gets compressed with PNG even though the image might be a JPG. There needs to be a distinction whether the image is JPG or PNG.

PR #809, opened 2020-12-27
PAINTROID-148: export as jpg + dialog

FromG:
Save image option:
If there is currently no image saved the dialog shows up with the standard values.(filename = "image", format = "jpg")
If there is a saved image the dialog will always show the saved image information.

Save copy option:
The copy option will always show the standard format. ("image5" "jpg");

when opened from catroid(e.g. pocketcode):
the dialog will not show unless you use the option export.

Overwriting:
When the user wants to overwrite an existing file a dialog pops up which will warn the user about it.
Export should work like save copy where the dialog will show the standard filename since it uses the same function as save copy,

Behavior:
Whenever you save an Image a Save dialog shows up where you can choose between
format and if you choose jpg you can then change the quality(also the transparent background will be changed to white when the picture gets saved). You can also tap
on the info button to get information about the file format.

**old branch thread: https://github.com/Catrobat/Paintroid/pull/789**
Reason for new branch:  git history cleanup. 

### DIFF
--- a/Paintroid/src/main/java/org/catrobat/paintroid/FileIO.java
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/FileIO.java
@@ -22,8 +22,11 @@ package org.catrobat.paintroid;
 import android.app.Activity;
 import android.content.ContentResolver;
 import android.content.ContentValues;
+import android.database.Cursor;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
+import android.graphics.Canvas;
+import android.graphics.Color;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Environment;
@@ -37,8 +40,6 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.text.SimpleDateFormat;
-import java.util.Date;
 import java.util.Locale;
 import java.util.Objects;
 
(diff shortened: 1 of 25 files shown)


Paintroid/src/main/java/org/catrobat/paintroid/FileIO.java
```diff
@@ -85,28 +104,34 @@ public static Uri saveBitmapToFile(String fileName, Bitmap bitmap, ContentResolv
 		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
 			ContentValues contentValues = new ContentValues();
 			contentValues.put(MediaStore.Images.Media.DISPLAY_NAME, fileName);
-			contentValues.put(MediaStore.Images.Media.MIME_TYPE, "image/png");
+			contentValues.put(MediaStore.Images.Media.MIME_TYPE, "image/*");
```
Electronix1337, line 107: If i try to edit a picture from catroid and export it with default name (e.g. image2, format="jpg") the resulting file is a .bin file with mimetype application/octet-stream. I'm not sure if this happens due to this line but i'd suggest to distinguish between png/jpeg and set the corresponding mimetype rather than image/*

Electronix1337 (changes requested):
Dialog is now correctly shown when user clicks "Export". The only thing that doesn't seem to work is the overwrite functionality when opened from catroid as can be seen in the attached picture. 
image
In Paintroid the file gets overwritten all the time, doesn't matter if the standard filename or user input is used which is not happening when opened from catroid. Rest lgtm.

Electronix1337 (approved):
Overwrite dialog doesn't show up anymore, lgtm.

wslany:
Thanks!!

PR #813, opened 2020-12-29
PAINTROID-156 spray can tool

StanglTobias:
- Ignored tests that fail because PAINTROID-126 did not adapt tests. Tests should be fixed in a separate ticket
- Added spray can tool with changeable radius.
- Removed Pipette tool from tool selection
- Failing tests pass on local device 

### DIFF
--- a/Paintroid/build.gradle
+++ b/Paintroid/build.gradle
@@ -117,6 +117,8 @@ dependencies {
 
     implementation 'androidx.core:core-ktx:1.3.2'
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.8'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.8'
 
     debugImplementation 'androidx.multidex:multidex:2.0.0'
 
--- a/Paintroid/gradle/code_quality_tasks.gradle
+++ b/Paintroid/gradle/code_quality_tasks.gradle
@@ -20,7 +20,7 @@
 task checkstyle(type: Checkstyle) {
     configFile file('config/checkstyle.xml')
     source '.'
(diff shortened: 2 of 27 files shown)


FromG (approved):
lgtm.

wslany (changes requested):
Very nice!

Four changes please:

1) The size of the spray area is too large, which makes it difficult to use it meaningfully in practice, see screenshot below (for an image of size 720 × 1480). The size numbers are written next to the spray area, and also a line drawn with the brush of the same width next to it. Please make it so that the width is **three times** the width of the corresponding brush size. Currently it seems more like a factor of 10 times for the 50 size, but even more for the 1 size.
Screenshot_20210109-092350

2) Make the starting value the same as for the brush, i.e., it shouldn't be 1 but 6 as the default value. Afterwards, it shall not have to be coordinated with the width of the brush, i.e., no change for that. This is just about the initial value. 
 
3) Make it spray, i.e., appear, a bit more slowly, so that the density appears more slowly, so one has more time to control it, _especially_ when drawing lines, which is a typical case. I would say, around 3 times more slowly. See the following screenshot on how fast it appears now, when one draws a line with the same, normal speed that allows to control where the line goes. Note that this is also linked with number 1 above and number 4 below. It should be coordinated and balanced with them, to make it easy to draw precisely, in a controlled way.
Screenshot_20210109-100334

4) Make it appear the slower the smaller the size, so that it looks approximately of the same density for the same duration, independently of the size. Here an image how it currently looks like, on the left, when I touched as shortly as possible, and on the right, when I touched a bit longer, approximately for the same amount of time for each size. The density should be roughly the same, but it isn't:
Screenshot_20210109-092459

wslany:
One more thing regarding the speed of spraying: it should be the same on future devices, i.e., not become faster the faster the hardware becomes.

FromG (approved):
lgtm

wslany (approved):

wslany:
Thanks!

PR #830, opened 2021-02-16
Translated some strings on russian

andrey-pr:


wslany:
Translations please via https://translate.catrobat.org/ only.

PR #862, opened 2021-04-03
PAINTROID-229 The brush tools alignment in right to left languages such as arabic

Edward1919:
I changed the alignment of brush tools in right to left languages to remain in the left left to right format.
Line of jira issue PAINTROID-229
https://jira.catrob.at/browse/PAINTROID-229

after
before


ThorstenBandel:
Hi,
everything looks good, i would have just one thing: could you add tests for this? Basically just to check if ltr is working

Edward1919:
@ThorstenBandel Dear sir, I added the test for RTL and is passed. Please review this. Thanks.

ThorstenBandel (approved):
Looks good to me

wslany (changes requested):
Thanks, looks good! However, I found something strange: the number entry fields contain strange cursors that are continuously blinking, at least on certain devices. This looks very nervous and could be distracting to users, and does not happen in the left-to-right mode of the app. Also, other text fields on the device do not blink in this way, so it must be something in our app that does something strange. I tried to capture in a screenshot, the indicated two cursor parts are blinking on and off around every 0.4 seconds.
image

It also happens at other similar places, e.g., here:
image

The strange thing is that it does not happen everywhere, for instance not here: 
image
or here:
image

Could it be because of the size of the entry field?

Also, I found several other layout issues related to right-to-left languages, but I think they should be better handled in another ticket, so I created https://jira.catrob.at/browse/PAINTROID-241

Edward1919:
@wslany Dear Sir please mention on what devices this occurs or on what device you noticed this behavior? It behaves normal on android 9.

wslany:
image

Edward1919:
@wslany 
Dear sir the changes you request is applied. 
Seems the strange behavior of blinking cursor happens when our edittext is wrap with a textinputlayout using material library.
P.S. The test is already available in BrushPickerIntegrationTest for edittext  and it passes those tests I don't think we need to add any extra test for it.
Please review this.
Thanks.

wslany:
Thanks! Please squash to one commit.

Edward1919:
Sir I have mistakenly reverted all commits, and could not squash commits I don't know why.
What is your suggestion. I was thinking of creating a new pull request and adding all changes in one commit confirm if you agree please.

wslany:
Yes please create a new PR, no problem.

wslany:
Followup in #871

PR #883, opened 2021-04-21
PAINTROID-237 fixed ActivityNotFound when rating the app

ldirry:
Fixed ActivityNotFoundException when the user tries to rate this app but no app is registered for handling the intent.

https://jira.catrob.at/browse/PAINTROID-237

### DIFF
--- a/Paintroid/src/main/java/org/catrobat/paintroid/ui/MainActivityNavigator.java
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/ui/MainActivityNavigator.java
@@ -22,6 +22,7 @@ package org.catrobat.paintroid.ui;
 import android.annotation.SuppressLint;
 import android.content.ActivityNotFoundException;
 import android.content.Intent;
+import android.content.pm.ActivityInfo;
 import android.content.pm.PackageManager;
 import android.database.Cursor;
 import android.graphics.Bitmap;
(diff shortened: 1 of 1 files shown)


ldirry:
Should a toast message be displayed when no app can be found for handling the intent? (They honestly feel a bit cheap to me but at least the user gets feedback)

ThorstenBandel (approved):
LGTM

wslany (changes requested):
I just tried this on a device without app store installed, and the app closed and I returned to the Android home screen when I tried to rate Pocket Paint. Unfortunately I did not get a stack trace. Here the details about the phone:
image

Same also on these devices:
image

image

I checked with the develop branch with these phones, and the behavior was exactly the same, so the PR seems not to have changed anything. What to do?

CodeChamp-SS (approved):
LGTM 👍🏻

wslany (approved):

wslany:
Thanks!

wslany:
> I also tested it now on browserstack with these devices. The problem seems to be that browserstack doesn't allow these app stores to be opened. If you try opening the AppGalery on Huawai or App Store on the oppo device, the stores will immedately get minimized. (Opening Google play store on the Pixel 4 (for example) seems to be allowed).
>
> But the pocket paint actually never crashes, when opening one of these stores, for example you can draw a line in pocket paint, then rate pocket paint (device will go back to home screen) and then you can reopen the app and the line is still there. On the oppo device you can kinda see the minimize animation, that gets triggered when the home screen button gets pressed, I assume this is how they handle the scenario when opening an app, you don't have permission for. (It all feels a bit buggy and laggy though).

Thanks, yeah I also saw that Browserstack is very laggy with these devices. I wrote to the browserstack support this message:
> Hi, we just found that we cannot access the Huawei AppGallery on the Huawei device. Same also on Oppo's app store. The app store apps immediately close when we open them. It would be nice to be able to check that the connection to them work via intent as we want to be able to see whether our app redirects to the correct app store depending on the device. Thanks and all the best, Wolfgang

wslany:
Answer from Browserstack support:
> Access to the default applications has been restricted on our Oppo, Huawei and Vivo devices at the moment, but our team is working towards adding support for the same.

PR #894, opened 2021-05-12
PAINTROID-161 Save/load in .catrobat-image format (2)

ldirry:
So I decided to go with the Kryo (https://github.com/EsotericSoftware/kryo) Library for Serialization since it was the library that according to different benchmarks and articles performed the best in terms of output size (e.g. https://github.com/eishay/jvm-serializers/wiki).  Also it was recommended on multiple stackoverflow posts.

I also wrote custom Serializers for every object that gets serialized, mainly to ensure backwards compatibility and minimum file size. The idea is that a Serializer specifies how an object should be written and read from a file.
Without custom Serializers you would usually experience some limitations like for instance not being able to change the type of a member variable, if you wanted to keep backwards compatibility.
All these custom Serializers inherit from the class `VersionSerializer`. With this class we can specify how we want to read an object from different Catrobat image file versions.

For example:
If we wanted to introduce a new feature with 3D Coordinates, then our `PointCommand` would not be able to take the `PointF` parameter anymore and would instead need a class that can store 3 coordinates, lets call it `Point3F`. All we know have to do is, increment the field `CommandSerializationUtilities.CURRENT_IMAGE_VERSION` and in the class `PointCommandSerializer`, we override the function `readV1(..)` and specify in this function how we convert the read `PointF` object from the old image version to the new `Point3F`. After that we just return a new `PointCommand` with the new `Point3F` as parameter.
Also in the `write(..)` function we specify that we want to write a `Point3F` instead and in the `readCurrentVersion(..)` we read the `Point3F`. 

If we want to introduce a new Command then we don't even have to increment the `CURRENT_IMAGE_VERSION`. Just create a new Serializer similar to the other ones and register it in the function `CommandSerializationUtilities.setRegisterMapVersion()`.
IMPORTANT: (I also added a comment in code),
Never change the order of these and only add/register new Serializers at the end of this function. Basically kryo assigns an ID to each class, this way it doesn't have to write/store the full package+name of each class.

Forwards compatibility is not supported (and would result in an image couldn't be loaded dialog), else we would also need to specify what happens when writing an object for every version. (which in my opinion is not worth the effort since users should just update to the newest version on the play store)

I also had to exchange the `android.graphics.Path` with a new class `SerializablePath` (inherits from Path). This class basically records/remembers all important function calls to this object. Else there was no option to get this information for serialization because `android.graphics.Path` seems to use some sort of registry system to maybe store some data more efficently for reuse.

Added following tests:
- CatrobatImageIOIntegrationTest.kt (to test the process of writing/reading a catrobat-image)
- CommandSerializationTest.kt (to test that every command is serialized and read as intended)

Also refactored all the Commands to Kotlin and added testcases for the commands which didn't have any before.
Just some thoughts on what I changed when refactoring the Commands to Kotlin.
- when calling `run(canvas: Canvas, layerModel: LayerContracts.Model)`, the canvas and layermodel should and are currently never null (because I had to remove a test that checked for null)
- completly refactored the StampCommand and removed BaseCommand, since the StampCommand was the only class extending it, and the BaseCommand had only functions that were useful for the StampCommand (like storing bitmap)
(all commands inherit from the `Command` Interface still)


ThorstenBandel:
https://confluence.catrob.at/display/PAINT/.catrobat+File+Format

ThorstenBandel (approved):
LGTM

wslany:
Thanks!

PR #917, opened 2021-06-04
PAINTROID-254 High frequency crash in paintroid.ui.dragndrop.DragAndDropListView.dispatchDraw

ldirry:
The problem would occur if a large image was loaded (>6000x6000 px) and when the user then tried to open the layer menu. The `super.dispatchDraw` in the `DragAndDropListView` would throw an exception if the bitmap was greater than 100mb (which seems to be roughly at 6000x6000). So to fix this problem, I greatly resized the bitmap/preview images in the layer menu. This also reduced lag in the layer menu for larger images that would work previously.

However, I would then sometimes get an outofmemory error, if I tried to add new layers (for large images >6000x6000), so I had to increase the scaleFactor (for new large images that are being loaded), since the calculation didn't seem to consider potential multiple layers.

I didn't cache the resized preview-images from the layermenu, since users usually change the image between each resize. The resizing starts in a coroutine when the user starts opening the layer menu.

refactored to Kotlin: `LayerAdapter.kt`, `LayerIntegrationTest.kt`

I had many runs on jenkins, because for some odd reasons, the added test: `testLoadLargeBitmapAndAddMaxLayerMenu`, didn't work on Jenkins. I had it in `BitmapIntegrationTest.java` before, tried different approaches on how to test loading an image with an `Intent`, all worked out locally but none on Jenkins. However, after I moved this function to `LayerIntegrationTest.kt` it worked... (both these files are in the same package though..)

https://jira.catrob.at/browse/PAINTROID-254


Paintroid/src/main/java/org/catrobat/paintroid/listener/DrawerLayoutListener.kt
```diff
@@ -0,0 +1,29 @@
...
+            for (i in 0 until adapter.count) {
+                adapter.getViewHolderAt(i)?.let { holder ->
+                    holder.bitmap?.let { bitmapImageView ->
+                        holder.updateImageView(bitmapImageView, true)
+                    }
+                }
+            }
+        }
+    }
+}
```
ThorstenBandel, line 29: Could you add a newline at the end of the file please

ThorstenBandel (approved):
LGTM

wslany:
Thanks!

PR #1002, opened 2021-08-14
PAINTROID-347 ANR in ColorPickerDialog

CodeChamp-SS:
The ANR occurs on clicking the `pipette` button when the image is very large due to the use of `runBlocking` which freezes the UI/main thread till the tasks get completed. `runBlocking` is almost never a tool you use in production. It undoes the asynchronous, non-blocking nature of coroutines. So replaced it with `CoroutineScope` in order to obtain the real asynchronous behaviour. As the `storeBitmapTemporally` method is now properly being performed in a background thread, so now the UI isn't frozen and the user could press the `pipette` button multiple times, so to prevent that I disable the button once the user presses it, so this also gives a hint to the user that the button has been pressed and also prevents the user from clicking the button multiple times(which would lead to the launching of `ColorPickerPreviewActivity` multiple times).
One small problem with the new implementation(which I'm not able to fix as of now) is that if the user changes the orientation while the `storeBitmapTemporally` is still running asynchronously, the app crashes. I've prevented the crash by catching the exception, but the user would need to press the `pipette` button once again in that case.

https://jira.catrob.at/browse/PAINTROID-347

### DIFF
--- a/colorpicker/src/main/java/org/catrobat/paintroid/colorpicker/ColorPickerDialog.kt
+++ b/colorpicker/src/main/java/org/catrobat/paintroid/colorpicker/ColorPickerDialog.kt
@@ -35,6 +35,7 @@ import android.graphics.drawable.ColorDrawable
 import android.graphics.drawable.Drawable
 import android.graphics.drawable.RippleDrawable
 import android.os.Bundle
+import android.util.Log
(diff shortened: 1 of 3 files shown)


ldirry (approved):
Looks nice to me. ^^

PR #1019, opened 2021-10-31
PAINTROID-324 When you load an image and select colour pipette app hangs for a while

barcodepony:
removing coroutines that save and load the image, resulting in freeze.

Previously coroutines were used to temporarily save and restore an image to pass it to the new intent which is started when we click the pipette inside the color picker. The reason most likely was the byte-limitation for extras for new intents.
A static variable was introduced where a bitmap can be passed to the class. If no bitmap is set it will fall back to a colorless 512x512 bitmap to keep the current behavior.

The freeze is now completely removed.

### DIFF
--- a/colorpicker/src/main/java/org/catrobat/paintroid/colorpicker/ColorPickerDialog.kt
+++ b/colorpicker/src/main/java/org/catrobat/paintroid/colorpicker/ColorPickerDialog.kt
@@ -56,7 +56,8 @@ private const val INITIAL_COLOR = "InitialColor"
 private const val REQUEST_CODE = 1
 private const val BITMAP_NAME = "temp.png"
 const val COLOR_EXTRA = "colorExtra"
-const val BITMAP_NAME_EXTRA = "bitmapNameExtra"
+const val BITMAP_HEIGHT_EXTRA = "bitmapHeightNameExtra"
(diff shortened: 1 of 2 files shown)


CodeChamp-SS (approved):
LGTM! The changes work as expected, the freeze is no longer there :).

ThorstenBandel:
Works great now,
One question / change request if its possible: Do we know when the picture is done loading  in the colorpicker? 
If so: could we add a loading spinner while its is loading the picture into the picker? When using larger pictures it needs a few seconds to load (for the first time after the picture is loaded in) and there is just a white image instead of the imported picture for these seconds. 

It's not a must have, but would be nice from UX view

barcodepony:
@ThorstenBandel I can not replicate the issue. The image in the color picker is displayed instantly. at least for me, as it should be.
The point of my implementation was to remove the loading time completely - and the view where we can pick a color should never be delayed or displayed empty. Therefore the color picker _should be done loading_ on creation before anything else happens. Are you merging any related issues that could be interfering?

ThorstenBandel:
Ok, it wasn't too much of a problem. Its a huge improvement even if it sometimes happens. If there are problems that arise we will create a new ticket, Thanks :)

PR #1033, opened 2022-01-07
PAINTROID-382: "+" icon not appearing in line tool after drawing the first line segment

JulianJautz:
"+" didn't not appear when line was drawn in one swipe
(https://jira.catrob.at/browse/PAINTROID-382)

* setting startpoint plus endpoint and drawing a line was handled differently
* drawing a line in one swipe behaves now like setting startpoint and endpoint separately
* also fixes:
   PAINTROID-383: Line tool does not allow to correctly reposition end point of first segment
   (https://jira.catrob.at/browse/PAINTROID-383)

### DIFF
--- a/Paintroid/src/main/java/org/catrobat/paintroid/tools/implementation/LineTool.kt
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/tools/implementation/LineTool.kt
@@ -32,12 +32,9 @@ import org.catrobat.paintroid.tools.ToolType
 import org.catrobat.paintroid.tools.Workspace
 import org.catrobat.paintroid.tools.common.CommonBrushChangedListener
 import org.catrobat.paintroid.tools.common.CommonBrushPreviewListener
-import org.catrobat.paintroid.tools.common.LINE_THRESHOLD
-import org.catrobat.paintroid.tools.implementation.LineTool.Companion.topBarViewHolder
 import org.catrobat.paintroid.tools.options.BrushToolOptionsView
 import org.catrobat.paintroid.tools.options.ToolOptionsVisibilityController
 import org.catrobat.paintroid.ui.viewholder.TopBarViewHolder
-import kotlin.math.abs
 
 class LineTool(
     private val brushToolOptionsView: BrushToolOptionsView,
@@ -84,6 +81,10 @@ class LineTool(
     var connectedLines = false
 
     var undoRecentlyClicked = false
+
+    var undoPreviousLineForConnectedLines = true
+
+    var changeInitialCoordinateForHandleNormalLine = false
 
     companion object {
         var topBarViewHolder: TopBarViewHolder? = null
@@ -146,9 +147,9 @@ class LineTool(
 
     fun onClickOnPlus() {
         if (startpointSet && endpointSet) {
-            val newStartCoordinate = endPointToDraw
-            initialEventCoordinate = endPointToDraw
-            previousEventCoordinate = endPointToDraw
+            val newStartCoordinate = endPointToDraw?.let { PointF(it.x, it.y) }
+            initialEventCoordinate = endPointToDraw?.let { PointF(it.x, it.y) }
+            previousEventCoordinate = endPointToDraw?.let { PointF(it.x, it.y) }
             startPointToDraw = null
             endPointToDraw = null
             startpointSet = false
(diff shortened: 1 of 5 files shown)


barcodepony (approved):
Changes and Tests look good to me, now works as intented! 👍 

Failed tests in CI/CD appeare to be flaky and can be ignored.

wslany (changes requested):
Thanks! However, the functionality is still not correct, as follows:

1. Select the line tool, then touch the screen at point A and, while continuing to touch the screen, move your finger around on the screen towards point B. 
2. Observe that, correctly, the first line is drawn starting from A to B and continuously updated to the latest touch position B.
3. Remove the finger from the screen, then tap at a new position C.
4. Observe that _incorrectly_ a second line segment is drawn from B to C. Instead, the expected behavior would be to update B of the first segment to the new end position C, so that there is a single line segment from A to C, since neither the checkmark nor the "+" button was touched, and the first segment (starting from A) has not yet been finished. 
5. Note that sometimes, the first line segment is deleted, and sometimes it remains. It is unclear what actions influence whether this first segment remains or disappears. 

Another, probably unrelated problem:

1. Using the line tool, draw a first segment from point A to point B by tapping on position A then on position B. Then tap on a third position C. 
2. Observe that, as expected, the segment is continuously updated from A to the latest point C. Tapping can be repeated several times.
3. Now touch the screen at another position, D, and, without removing the finger from the screen, move the finger to another position E. 
4. Observe that during this movement (while the finger still touches the screen), a second segment is _incorrectly_ drawn between points D and E. Instead, the expected behavior should have been to update the first segment from A to E, as in the first step, since neither "+" nor the checkmark have yet been tapped, and the first segment was not yet finalized. 

Another issue, though I guess this one is probably the same as the one above:

1. Draw a first line segment from point A to B, then tap on "+".
2. Touch the screen at another position C and move your finger to position D, without removing the finger from the screen.
3. Observe that during this movement, a line is _incorrectly_ drawn and continuously updated from C to D. Instead, the correct, expected behavior should be that when the screen is touched initially at point C in step 2 above, the second segment is drawn from B to C, and then continuously updated starting from B to the finger's latest position, until the finger is removed from the screen.

wslany:
> Changes and Tests look good to me, now works as intented! 👍
> 
> Failed tests in CI/CD appeare to be flaky and can be ignored.

@barcodepony your reviews somehow are not accepted as reviews by github. Maybe add an explicit positive review to make it stick? See below how to do that:
image

wslany:
@barcodepony Are the changes fine from a code-review point of view? I checked the functionality (including pinch zoom, undo redo, and switch of tools or colors), and from that perspective everything works fine now (@JulianJautz , awesome work!), but I didn't check test coverage etc. Please @barcodepony , can you do one (hopefully last) code-review round? 

🙏 Please also make sure to tap on "approve" in case you approve so that your approval is registered by the system.

barcodepony:
@wslany the code works implements the behaviour now correctly and is also properly tested.
Personally, I just don't think that the new variables in `LineTool.kt` are a pretty solution. Especially `fromMove` is not very self explanatory. From a technical point of view the PR is fine, unclear if is should approve this in regard of the "code style" or if it's better if @JulianJautz can find a better name for one or both variables in question. Looking at the code in e.g. 1 year definitely could raise some questions.

@JulianJautz could you rename the variables quickly to resolve the issue? :+1:

Paintroid/src/main/java/org/catrobat/paintroid/tools/implementation/LineTool.kt
```diff
@@ -85,6 +82,10 @@ class LineTool(
 
     var undoRecentlyClicked = false
 
+    var firstMove = true
+
+    var fromMove = false
```
barcodepony, line 87: i suggest renaming these to variables to make there functionality more clear, especially `fromMove` does not give a lot of insight. `firstMove` could be renamed aswell to make it obvious that it's used for connected lines :smiley:

barcodepony (changes requested):
Please see my comment for `LineTook.kt` (:

JulianJautz:
> @wslany the code works implements the behaviour now correctly and is also properly tested. Personally, I just don't think that the new variables in `LineTool.kt` are a pretty solution. Especially `fromMove` is not very self explanatory. From a technical point of view the PR is fine, unclear if is should approve this in regard of the "code style" or if it's better if @JulianJautz can find a better name for one or both variables in question. Looking at the code in e.g. 1 year definitely could raise some questions.
> 
> @JulianJautz could you rename the variables quickly to resolve the issue? 👍

@barcodepony I changed the names now, but to really understand the behaviour, it would need more than a better name, I think. Should I add comments? And would you have had an idea for a different approach as I can imagine that there are better solutions. But I couldn't figure out another way to differentiate between taps and swipes, which is necessary already while drawing...

barcodepony:
> > @wslany the code works implements the behaviour now correctly and is also properly tested. Personally, I just don't think that the new variables in `LineTool.kt` are a pretty solution. Especially `fromMove` is not very self explanatory. From a technical point of view the PR is fine, unclear if is should approve this in regard of the "code style" or if it's better if @JulianJautz can find a better name for one or both variables in question. Looking at the code in e.g. 1 year definitely could raise some questions.
> > @JulianJautz could you rename the variables quickly to resolve the issue? 👍
> 
> @barcodepony I changed the names now, but to really understand the behaviour, it would need more than a better name, I think. Should I add comments? And would you have had an idea for a different approach as I can imagine that there are better solutions. But I couldn't figure out another way to differentiate between taps and swipes, which is necessary already while drawing...

@JulianJautz to the best of my knowledge, comments are undesired in catrobat projects and the code should speak for itself.
The new names are way better and I think we can now leave it like that with the new changes. I managed to figured out the code behaviour with the old names as well, so it should be good enough to approve the changes.

Regarding a different ideas: I do not see any obvious alternative approaches, my main concern is that we need two new state variables. What could be done is a refactor with e.g. a state enum for better indication of what a certain function should do in the given situation - but a refactor like this outside the responsibilites of the ticket.

Changes approved! 😄👍

barcodepony (approved):

wslany (approved):

wslany:
Thanks!!!! Super!

PR #1034, opened 2022-01-27
PAINTROID-371 IndexOutOfBoundsException paintroid.model.LayerModel.getLayerAt and similar crashes with layers (#3 in Crashes and ANRs)

barcodepony:
Added error handling to the LayerModel for invalid index accesses. Errors occured due to concurrency issues - now null types are returned instead of crashing.  All major occurences now have null type handling in place, resulting in silent fails for the relevant commands, leaving a valid state behind.  Command undo's are no handled. Missing tests for the LayerModel have been added.

### DIFF
--- a/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/junit/model/LayerTest.java
+++ b/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/junit/model/LayerTest.java
@@ -1,6 +1,6 @@
 /*
  * Paintroid: An image manipulation application for Android.
- * Copyright (C) 2010-2015 The Catrobat Team
+ * Copyright (C) 2010-2022 The Catrobat Team
  * (<http://developer.catrobat.org/credits>)
  *
  * This program is free software: you can redistribute it and/or modify
@@ -39,7 +39,11 @@ import org.junit.runner.RunWith;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
 import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.timeout;
@@ -50,14 +54,16 @@ public class LayerTest {
 	private CommandFactory commandFactory;
 	private CommandManager commandManager;
 	private LayerContracts.Model layerModel;
+	private final int layerHeight = 200;
+	private final int layerWidth = 200;
 
 	@Before
 	public void setUp() {
 		commandFactory = new DefaultCommandFactory();
 		layerModel = new LayerModel();
-		layerModel.setWidth(200);
-		layerModel.setHeight(200);
-		Layer layer = new Layer(Bitmap.createBitmap(200, 200, Bitmap.Config.ARGB_8888));
+		layerModel.setWidth(layerWidth);
+		layerModel.setHeight(layerHeight);
+		Layer layer = new Layer(Bitmap.createBitmap(layerWidth, layerHeight, Bitmap.Config.ARGB_8888));
 		layerModel.addLayerAt(0, layer);
 		layerModel.setCurrentLayer(layer);
 
@@ -154,5 +160,33 @@ public class LayerTest {
 		assertThat(layerToHide.getBitmap().getPixel(2, 1), is(Color.BLACK));
 		assertThat(layerToHide.getBitmap().getPixel(3, 1), is(Color.BLACK));
 		assertThat(layerToHide.getBitmap().getPixel(4, 1), is(Color.BLACK));
+	}
+
+	@Test
+	public void testGetLayerAt() {
+		for (int i = 0; i < layerModel.getLayerCount(); i++) {
+			assertNotNull(layerModel.getLayerAt(i));
+		}
+
+		assertNull(layerModel.getLayerAt(-1));
+		assertNull(layerModel.getLayerAt(layerModel.getLayerCount()));
+	}
+
+	@Test
+	public void testAddLayerAt() {
+		Layer layer = new Layer(Bitmap.createBitmap(layerWidth, layerHeight, Bitmap.Config.ARGB_8888));
+		assertTrue(layerModel.addLayerAt(0, layer));
+		assertTrue(layerModel.addLayerAt(layerModel.getLayerCount(), layer));
+		assertFalse(layerModel.addLayerAt(-1, layer));
+		assertFalse(layerModel.addLayerAt(layerModel.getLayerCount() + 1, layer));
+	}
+
+	@Test
+	public void testRemoveLayerAt() {
+		assertFalse(layerModel.removeLayerAt(-1));
+		assertTrue(layerModel.removeLayerAt(0));
+		Layer layer = new Layer(Bitmap.createBitmap(layerWidth, layerHeight, Bitmap.Config.ARGB_8888));
+		layerModel.addLayerAt(0, layer);
+		assertFalse(layerModel.removeLayerAt(layerModel.getLayerCount()));
 	}
 }
--- a/Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
@@ -1,6 +1,6 @@
 /*
  * Paintroid: An image manipulation application for Android.
- * Copyright (C) 2010-2021 The Catrobat Team
+ * Copyright (C) 2010-2022 The Catrobat Team
  * (<http://developer.catrobat.org/credits>)
  *
  * This program is free software: you can redistribute it and/or modify
@@ -90,23 +90,37 @@ class DefaultCommandManager(
     }
 
     override fun undo() {
+        var success = true
         val command = undoCommandList.pop()
         redoCommandList.addFirst(command)
-
         var layerCount = layerModel.layerCount
         val currentCommandName = command.javaClass.simpleName
         val addLayerCommandRegex = AddLayerCommand::class.java.simpleName.toRegex()
         val mergeLayerCommandRegex = MergeLayersCommand::class.java.simpleName.toRegex()
 
+        var backupLayer: LayerContracts.Layer? = null
         if (currentCommandName.matches(addLayerCommandRegex)) {
             layerCount--
-            layerModel.removeLayerAt(0)
+            backupLayer = layerModel.getLayerAt(0)
+            success = layerModel.removeLayerAt(0)
         }
 
         val checkBoxes: MutableList<Boolean> = ArrayList(Collections.nCopies(layerCount, true))
 
         if (!currentCommandName.matches(mergeLayerCommandRegex)) {
-            backUpCheckBoxes(layerCount, checkBoxes)
+            success = if (!backUpCheckBoxes(layerCount, checkBoxes)) false else success
+        }
+
+        if (!success) {
+            backupLayer?.apply {
+                if (layerModel.getLayerIndexOf(this) == -1) {
+                    layerModel.addLayerAt(0, this)
+                }
+            }
+            val commandToRestore = redoCommandList.pop()
+            undoCommandList.addFirst(commandToRestore)
+
+            return
         }
 
         layerModel.reset()
@@ -172,14 +186,21 @@ class DefaultCommandManager(
         }
     }
 
-    private fun backUpCheckBoxes(layerCount: Int, checkBoxes: MutableList<Boolean>) {
+    private fun backUpCheckBoxes(layerCount: Int, checkBoxes: MutableList<Boolean>): Boolean {
+        var success = true
         if (layerCount > 1) {
             for (index in layerCount - 1 downTo 0) {
-                checkBoxes[index] = layerModel.getLayerAt(index).checkBox
+                layerModel.getLayerAt(index)?.let {
+                    checkBoxes[index] = it.checkBox
+                } ?: run { success = false }
             }
         } else {
-            checkBoxes[0] = layerModel.getLayerAt(0).checkBox
+            layerModel.getLayerAt(0)?.let {
+                checkBoxes[0] = it.checkBox
+            } ?: run { success = false }
         }
+
+        return success
     }
 
     private fun fetchBackCheckBoxes(layerCount: Int, checkBoxes: List<Boolean>) {
@@ -187,8 +208,10 @@ class DefaultCommandManager(
             for (index in layerCount - 1 downTo 0) {
                 val destinationLayer = layerModel.getLayerAt(index)
                 if (!checkBoxes[index]) {
-                    destinationLayer.switchBitmaps(false)
-                    destinationLayer.checkBox = false
+                    destinationLayer?.let {
+                        it.switchBitmaps(false)
+                        it.checkBox = false
+                    }
                 }
             }
         } else {
--- a/Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/MergeLayersCommand.kt
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/MergeLayersCommand.kt
@@ -1,6 +1,6 @@
 /*
  * Paintroid: An image manipulation application for Android.
- * Copyright (C) 2010-2021 The Catrobat Team
+ * Copyright (C) 2010-2022 The Catrobat Team
  * (<http://developer.catrobat.org/credits>)
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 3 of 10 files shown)


Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
```diff
@@ -175,10 +175,10 @@ class DefaultCommandManager(
     private fun backUpCheckBoxes(layerCount: Int, checkBoxes: MutableList<Boolean>) {
         if (layerCount > 1) {
             for (index in layerCount - 1 downTo 0) {
-                checkBoxes[index] = layerModel.getLayerAt(index).checkBox
+                checkBoxes[index] = layerModel.getLayerAt(index)!!.checkBox
```
CodeChamp-SS, line 178: `!!` should never be used in production code, because it can result in unnecessary crashes.
Instead you can write something like this `checkBoxes[index] = layerModel.getLayerAt(index)?.checkBox == true`

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
```diff
@@ -175,10 +175,10 @@ class DefaultCommandManager(
...
             }
         } else {
-            checkBoxes[0] = layerModel.getLayerAt(0).checkBox
+            checkBoxes[0] = layerModel.getLayerAt(0)!!.checkBox
```
CodeChamp-SS, line 181: Please refer to changes suggested in line 178

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/MergeLayersCommand.kt
```diff
@@ -28,18 +29,31 @@ class MergeLayersCommand(position: Int, mergeWith: Int) : Command {
...
+        var success = false
+        if (sourceLayer != null && destinationLayer != null) {
+            val destinationBitmap = destinationLayer.bitmap
+            destinationBitmap ?: return
+            val copyBitmap = destinationBitmap.copy(destinationBitmap.config, true)
+            val copyCanvas = Canvas(copyBitmap)
+            copyCanvas.drawBitmap(sourceLayer.bitmap ?: return, 0f, 0f, null)
+            if (layerModel.removeLayerAt(position)) {
+                destinationLayer.bitmap = copyBitmap
+                if (sourceLayer === layerModel.currentLayer) {
```
CodeChamp-SS, line 48: you can simply write `==` instead of `===`

Paintroid/src/main/java/org/catrobat/paintroid/model/LayerModel.kt
```diff
@@ -22,25 +22,32 @@ import android.graphics.Bitmap
...
+    override val layers: ArrayList<LayerContracts.Layer> = arrayListOf()
     override val layerCount: Int
         get() = layers.size
 
     override fun reset() {
         layers.clear()
     }
 
-    override fun getLayerAt(index: Int): LayerContracts.Layer = layers[index]
+    override fun getLayerAt(index: Int): LayerContracts.Layer? = layers.getOrNull(index)
```
CodeChamp-SS, line 39: ah.. very nice :)

Paintroid/src/main/java/org/catrobat/paintroid/model/LayerModel.kt
```diff
@@ -22,25 +22,32 @@ import android.graphics.Bitmap
...
-        layers.add(index, layer)
+    @SuppressWarnings("SwallowedException", "TooGenericExceptionCaught")
+    override fun addLayerAt(index: Int, layer: LayerContracts.Layer): Boolean {
+        return try {
+            layers.add(index, layer)
+            true
+        } catch (e: IndexOutOfBoundsException) {
+            false
+        }
     }
```
CodeChamp-SS, line 51:
```suggestion
    @SuppressWarnings("SwallowedException", "TooGenericExceptionCaught")
    override fun addLayerAt(index: Int, layer: LayerContracts.Layer): Boolean = try {
        layers.add(index, layer)
        true
    } catch (e: IndexOutOfBoundsException) {
        false
    }
```
You can use expression body syntax here

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
```diff
@@ -187,7 +187,7 @@ class DefaultCommandManager(
             for (index in layerCount - 1 downTo 0) {
                 val destinationLayer = layerModel.getLayerAt(index)
                 if (!checkBoxes[index]) {
```
CodeChamp-SS, line 189:
```suggestion
                if (destinationLayer != null && !checkBoxes[0]) {
                destinationLayer.switchBitmaps(false)
                destinationLayer.checkBox = false
            }
```
You can remove `!!` and write something like suggested above :)
Check the if block in line 196 of this file.

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/RemoveLayerCommand.kt
```diff
@@ -28,8 +28,9 @@ class RemoveLayerCommand(position: Int) : Command {
     var position = position; private set
 
     override fun run(canvas: Canvas, layerModel: LayerContracts.Model) {
-        layerModel.removeLayerAt(position)
-        layerModel.currentLayer = layerModel.layers[0]
+        if (layerModel.removeLayerAt(position)) {
```
CodeChamp-SS, line 31: why this `if` block? 🤔

Paintroid/src/main/java/org/catrobat/paintroid/model/LayerModel.kt
```diff
@@ -50,8 +57,14 @@ class LayerModel : LayerContracts.Model {
         layers[position] = layer
     }
 
-    override fun removeLayerAt(position: Int) {
-        layers.removeAt(position)
+    @SuppressWarnings("SwallowedException", "TooGenericExceptionCaught")
+    override fun removeLayerAt(position: Int): Boolean {
```
CodeChamp-SS, line 61: refer changes suggested to line 44

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/junit/model/LayerTest.java
```diff
@@ -50,14 +54,16 @@
 	private CommandFactory commandFactory;
 	private CommandManager commandManager;
 	private LayerContracts.Model layerModel;
+	private final int layerHeight = 200;
```
CodeChamp-SS, line 57: nice!

Paintroid/src/main/java/org/catrobat/paintroid/presenter/LayerPresenter.kt
```diff
@@ -167,21 +167,23 @@ class LayerPresenter(
...
     override fun unhideLayer(position: Int, viewHolder: LayerContracts.LayerViewHolder) {
-        val destinationLayer = getDestinationLayer(position, true)
-        viewHolder.updateImageView(destinationLayer.bitmap, true)
-        drawingSurface?.refreshDrawingSurface()
-        if (model.currentLayer == destinationLayer) {
-            defaultToolController?.switchTool(ToolType.BRUSH, false)
-            bottomNavigationViewHolder?.showCurrentTool(ToolType.BRUSH)
+        getDestinationLayer(position, true)?.let { layer ->
+            viewHolder.updateImageView(layer.bitmap, true)
+            drawingSurface?.refreshDrawingSurface()
```
CodeChamp-SS, line 182: check suggestion at line 170

Paintroid/src/main/java/org/catrobat/paintroid/presenter/LayerPresenter.kt
```diff
@@ -167,21 +167,23 @@ class LayerPresenter(
     }
 
     override fun hideLayer(position: Int) {
-        val destinationLayer = getDestinationLayer(position, false)
-        drawingSurface?.refreshDrawingSurface()
-        if (model.currentLayer == destinationLayer) {
-            defaultToolController?.switchTool(ToolType.HAND, false)
-            bottomNavigationViewHolder?.showCurrentTool(ToolType.HAND)
+        getDestinationLayer(position, false)?.let { layer ->
+            drawingSurface?.refreshDrawingSurface()
```
CodeChamp-SS, line 171:
```suggestion
        drawingSurface?.refreshDrawingSurface()
        getDestinationLayer(position, false)?.let { layer ->
```
shouldn't `drawingSurface?.refreshDrawingSurface()` be called irrespective of whether `getDestinationLayer(position, false)` is null or not, the logic seems to be changed here 🤔, not very sure though

CodeChamp-SS (changes requested):
Apart from the changes suggested, the rest of the code changes LGTM. The approach to fix the ticket seems fine and should most probably(as I don't exactly know how to reproduce the crash) fix this crash.

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/RemoveLayerCommand.kt
barcodepony: removeLayerAt can fail and will retrung `false` . If it fails, we do not want to update the `currentLayer`

Paintroid/src/main/java/org/catrobat/paintroid/model/LayerModel.kt
barcodepony: good idea, missed it

Paintroid/src/main/java/org/catrobat/paintroid/model/LayerModel.kt
barcodepony: fixed

Paintroid/src/main/java/org/catrobat/paintroid/presenter/LayerPresenter.kt
barcodepony: no, we do not want to do anything here i think. We could not hide the layer since we could not fetch the destination layer (for some unknown concurrency reason) and therefore we do not want to interact and hope the user clicks again and hope that the (unknown) reason does not occure again.

Do you aggree? Not 100% sure either, defintely had to think about this.

Paintroid/src/main/java/org/catrobat/paintroid/presenter/LayerPresenter.kt
barcodepony: will also refer to the discussion at line 170

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
barcodepony: i now added a way to "recover" the undo/redo stack

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/MergeLayersCommand.kt
barcodepony: `===` was introduced by a previous commit and i would leave it there. It should be the same reference in this case and a referential equality compare is faster than a structural. I don't know the specific reason but it should not harm and would leave it as it is 😃

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
barcodepony: I now added a way to restore the case of a failure undoing the addLayerCommand and mergeLayerCommand.
AddLayerCommand gets restored completely, mergeLayerCommand should work as intented, if concurrency-issues occure, a layer which should be hidden COULD be set visible, due to default values being set in an error case.
Hope you think as well, that this is sufficient.

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/MergeLayersCommand.kt
CodeChamp-SS: yeah I know it was introduced in a previous commit, I was just doubtful whether these two values would always point to the same object when they're structurally equal 🤔. If you think they'll be, then it's fine :)

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/RemoveLayerCommand.kt
CodeChamp-SS: hmm.. yeah ok 👍🏻

Paintroid/src/main/java/org/catrobat/paintroid/model/LayerModel.kt
```diff
@@ -50,8 +55,12 @@ class LayerModel : LayerContracts.Model {
...
-    override fun removeLayerAt(position: Int) {
+    @SuppressWarnings("SwallowedException", "TooGenericExceptionCaught")
+    override fun removeLayerAt(position: Int): Boolean = try {
         layers.removeAt(position)
+        true
+    } catch (e: IndexOutOfBoundsException) {
+        false
     }
 
     companion object {
```
CodeChamp-SS, line 66: just noticed, could you please also remove the `@JvmStatic` annotations from the methods in the companion object, they're no longer required, they were added initially when some of the usages were in java.

CodeChamp-SS:
Just noticed, could you please also change the year in the copyright section of the LayerTest file to 2022?

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
CodeChamp-SS: The changes look good :)

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
```diff
@@ -90,23 +90,37 @@ class DefaultCommandManager(
...
         val checkBoxes: MutableList<Boolean> = ArrayList(Collections.nCopies(layerCount, true))
 
         if (!currentCommandName.matches(mergeLayerCommandRegex)) {
-            backUpCheckBoxes(layerCount, checkBoxes)
+            success = if (!backUpCheckBoxes(layerCount, checkBoxes)) false else success
+        }
+
+        if (!success) {
+            backupLayer?.apply {
+                if (layerModel.getLayerIndexOf(this) != -1) {
```
CodeChamp-SS, line 116: ah.. you're trying to check if the backupLayer is not present in the layerModel.layers, then you're adding it to the layerModel.layers, right 🤔? If yes, then I think the condition should be `layerModel.getLayerIndexOf(this) == -1`

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/MergeLayersCommand.kt
barcodepony: Since we both have our doubts, lets change it to `==`! :)

CodeChamp-SS: thanks, it would be better ig :)

Paintroid/src/main/java/org/catrobat/paintroid/presenter/LayerPresenter.kt
CodeChamp-SS: hmm.. makes sense 👍🏻. @ThorstenBandel could you please just take a look at it once, I think you'll have a better idea about this?

barcodepony: i know changed it according to your suggestion, shouldnt matter either way - but just in case ht drawingSurface receives a lock from the outside, it will be cleared in case of failure.

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/DefaultCommandManager.kt
barcodepony: Good catch! The backup layer could get popped from the list above, if it was removed, we want to add it back to the original position at position 0!

barcodepony:
> Apart from the changes suggested, the rest of the code changes LGTM. The approach to fix the ticket seems fine and should most probably(as I don't exactly know how to reproduce the crash) fix this crash.

I considered all requested changes and implemented them accordingly. Changes in question from the conversations above were changed to rule out any possible errors. Thanks for the very detailed review!

CodeChamp-SS:
> > Apart from the changes suggested, the rest of the code changes LGTM. The approach to fix the ticket seems fine and should most probably(as I don't exactly know how to reproduce the crash) fix this crash.
> 
> I considered all requested changes and implemented them accordingly. Changes in question from the conversations above were changed to rule out any possible errors. Thanks for the very detailed review!

Thanks for making all the suggested changes! The code LGTM, I'll approve the PR once the Jenkins run finishes :)

PR #1063, opened 2022-03-29
PAINTROID-249: Remove most regional codes from CrowdIn translation

AneiMakovec:
Renamed 'values-...' folders in Paintroid/src/main/res and colorpicker/src/main/res according to their names in Catroid.
Changed the crowdin_Paintroid.yml and crowdin_colorpicker.yml, by adding language mappings for android codes as it is done in Catroid.

https://jira.catrob.at/browse/PAINTROID-249


JulianJautz (approved):
LGTM
compared it to Catroid and it is implemented in the same way now. However, I don't completely understand all the naming conventions in Catroid (e.g. why is "pt-rPT" changed to solely pt (although there is still pt_BR), while all three "en"-Versions keep their additional identifier? Maybe a senior can elaborate on that?

tested several languages on the emulator

wslany:
Please resolve the conflicts (and sorry for my late review!!!)

> LGTM compared it to Catroid and it is implemented in the same way now. However, I don't completely understand all the naming conventions in Catroid (e.g. why is "pt-rPT" changed to solely pt (although there is still pt_BR), while all three "en"-Versions keep their additional identifier? Maybe a senior can elaborate on that?
> 
> tested several languages on the emulator

Reason: pt is the fallback for all regional variants of Portuguese (for instance in Macao), with one exception, pt_BR for Brazil, which reportedly is really quite different from the European variant of Portuguese. The population of Brazil is much much bigger, so in this special case it makes sense. Similar for Chinese --- modern Chinese (Mainland China) and traditional Chinese (Taiwan, but also Hongkong) are totally different, even use different font sets. 

This is not at all the case for instance for French in France versus French in Canada. Same for Spanish in all its countries, or Arabic. No big differences between the regional variants for these other languages apparently. 

Now for English, the situation again is different: EN without extension is US English for whatever legacy reasons for Android --- mostly because Google is a US company. The UK and all other regions from the former colonies of the UK use however the UK variant. Are there differences? Yes, for instance color (US) versus colour (everywhere else). But EN is the default... So, unfortunately, we need regional variants for significant other regions, like the UK, but also Australia, India, Canada.... it's...complex 😣

JulianJautz:
Thank you for the explanation! :)

AneiMakovec:
Thank you! Will look into it as soon as possible!

wslany:
🙏 @JulianJautz Please also check manually compiled with Catroid (via jenkins, build manually with Catroid checkbox ticked), changing the languages in the settings menu of Catroid and checking the correct language being displayed in the paint part for special case languages (pt, pr_BR,traditional and modern Chinese, and en versus en_AU, and report here whether successful or whether there are still problems. Check correctness by checking differences in the string.xml files.

AneiMakovec:
@JulianJautz I think all the conflicts are resolved, so it's ready for review :D

JulianJautz:
@wslany 
I tested the languages with Catroid build. The varieties of en, pt and chinese should work as intended. 
But I don't get the Serbian language options. We have three options in Pocket Code. I understand that there is a cyrillic and a latin alphabet version. And I assume that the corresponding language files are sr-rSP and sr_rCS? But what is the third option for (the middle one without parantheses)? Especially because it uses cyrillic in Pocket Code and latin in Pocket Paint. Also if I compare the cyrillic version in Pocket Paint with the sr-rSP language file we don't get all translations. E.g. there should be translations for "green" and "blue" in the color picker, but the english word is shown.

wslany:
OK, I just checked it manually compiled with Catroid for special case languages (pt, pr_BR,traditional and modern Chinese, and en versus en_AU), and it all worked fine. Awesome!

wslany:
Correct, Serbian is another special case, I forgot about it. So, we need another round of corrections for it. Sorry, @JulianJautz  --- I apologize for the premature merging. Shall we revert the PR, or fix it in another ticket/pr?

JulianJautz:
I think it is fine to leave this PR as it is and fix the Serbian language later on. :)

wslany:
Thanks! I just created https://jira.catrob.at/browse/PAINTROID-422 for Serbian.

PR #1067, opened 2022-04-11
PAINTROID-304 Text field size is reset if font or font-effect (bold, italic, ...) is changed

Electronix1337:
After this ticket the texttoolbox will not reset when font or font-effect is changed. Had to change some tests in the TextToolIntegrationTest suite. 

Additionally I did find a bug/unexpected behaviour where the text will be displayed slightly outside of the text tool box which I also tried to fix in this ticket but not sure if the result is perfect. My approach was to adjust the boxwidth when italic is applied.

https://jira.catrob.at/browse/PAINTROID-304


JulianJautz (dismissed):
As far as I understand the description in the ticket, the goal is that the size of the textbox doesn't change after the font was set to bold, italic,... (and not the font size itself). This is not what your implementation does.
Do I miss something?

And I also noticed that this change in size also happens, when the font is changed or "text" in the bottom bar is clicked, so always when something in the text tool related menu is changed.

Electronix1337:
> As far as I understand the description in the ticket, the goal is that the size of the textbox doesn't change after the font was set to bold, italic,... (and not the font size itself). This is not what your implementation does. Do I miss something?
> 
> And I also noticed that this change in size also happens, when the font is changed or "text" in the bottom bar is clicked, so always when something in the text tool related menu is changed.

You are right, but I talked to Thorsten before I started implementing the ticket and we both came to the same conclusion that the only reasonable solution to this would be to automatically adjust the font size when text box is resized using corner points. As the changes in font type and text style are applied to the text with the internal font size I would have to calculate the proportional difference when resizing but this seems like an overkill to me or rather counterintuitive in comparison to my implementation. I also don't see a proper use case where users want to distort the text field only in X or Y (so all resizing points except the corners) and then change the font type or text style as a distorted text is rarely the desired outcome. So in short, if a user resizes the text box and applies a change in font type or text style the desired outcome would be that the text box stays the same in size, which happens with my implementation and for this to properly work, I need to internally adjust the font size as the changes are made based on this. Hope it's clear what I mean with this.

JulianJautz:
I think I understand. So resizing via the corners seems to work as you described.
And the text box shrinkage after setting bold, italic,... won't be changed (when resizing box before only in x or y direction)?

> > And I also noticed that this change in size also happens, when the font is changed or "text" in the bottom bar is clicked, so always when something in the text tool related menu is changed.
> 
> I don't really get what you mean with this, could you further describe that please.
I meant that the box also shrinks when clicking on "Tools" or "Text" in the footer (box size change in x or y direction before). I think it would be good to not change the text box size when clicking these buttons.

Paintroid/src/main/java/org/catrobat/paintroid/tools/implementation/TextTool.kt
```diff
@@ -97,9 +99,14 @@ class TextTool(
...
 
-    private var textSize = DEFAULT_TEXT_SIZE
+    var textSize = DEFAULT_TEXT_SIZE
     private val stc: Typeface?
     private val dubai: Typeface?
+    var callback: TextToolOptionsView.Callback? = null
+    var oldBoxWidth = 0f
+    var oldBoxHeight = 0f
+    var oldToolPosition: PointF? = null
+    var wasItalicRecentlyDisabled = false
```
StanglTobias, line 109: can be removed

StanglTobias (changes requested):
There are still some problems when the text is italic

Screenshot_20220622_171909

StanglTobias (approved):

Paintroid/src/main/java/org/catrobat/paintroid/tools/implementation/TextTool.kt
```diff
@@ -198,15 +225,25 @@ class TextTool(
...
+        } else {
+            multilineText.forEachIndexed { index, textLine ->
+                canvas.drawText(
+                    textLine,
+                    -(scaledBoxWidth / 2) + scaledWidthOffset,
+                    -(scaledBoxHeight / 2) + scaledHeightOffset - textAscent + lineHeight * index,
+                    textPaint
+                )
+            }
         }
```
CodeChamp-SS, line 246: You can write this also as suggested in `TextToolCommand.kt`.

Paintroid/src/main/java/org/catrobat/paintroid/tools/implementation/TextTool.kt
```diff
@@ -218,11 +255,43 @@ class TextTool(
...
+
+    private fun applyAttributes() {
+        boxWidth = oldBoxWidth
+        boxHeight = oldBoxHeight
+        if (oldToolPosition != null) {
+            toolPosition = oldToolPosition as PointF
+        } else {
+            resetBoxPosition()
+        }
+    }
```
CodeChamp-SS, line 293:
```suggestion
    private fun storeAttributes(italic: Boolean = false) {
        if (italic) {
            boxWidth *= ITALIC_FONT_BOX_ADJUSTMENT
        }
        oldBoxWidth = boxWidth
        oldBoxHeight = boxHeight
        oldToolPosition = PointF(toolPosition.x, toolPosition.y)
    }

    private fun applyAttributes(italic: Boolean = false) {
        boxWidth = oldBoxWidth / if (italic) ITALIC_FONT_BOX_ADJUSTMENT else 1f
        boxHeight = oldBoxHeight
        if (oldToolPosition != null) {
            toolPosition = oldToolPosition as PointF
        } else {
            resetBoxPosition()
        }
    }
```
You can write it like this, to remove duplicity and clean up the code.

CodeChamp-SS (changes requested):
The changes work as expected. Please look into the suggested changes (just for cleaning up the code a bit). Then I'll approve the PR.

Paintroid/src/main/java/org/catrobat/paintroid/command/implementation/TextToolCommand.kt
```diff
@@ -69,13 +74,24 @@ class TextToolCommand(
...
+                }
+            } else {
+                multilineText.forEachIndexed { index, textLine ->
+                    canvas.drawText(
+                        textLine,
+                        -(scaledBoxWidth / 2) + scaledWidthOffset,
+                        -(scaledBoxHeight / 2) + scaledHeightOffset - textAscent + lineHeight * index,
+                        textPaint
+                    )
+                }
```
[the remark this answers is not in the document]
CodeChamp-SS: I think you also remove line 77 and use that directly inside the `drawText()` method in place of that `x`.

CodeChamp-SS (approved):
LGTM!

Paintroid/src/main/java/org/catrobat/paintroid/tools/implementation/TextTool.kt
```diff
@@ -100,9 +102,13 @@ class TextTool(
     @JvmField
     var bold = false
 
-    private var textSize = DEFAULT_TEXT_SIZE
+    var textSize = DEFAULT_TEXT_SIZE
     private val stc: Typeface?
     private val dubai: Typeface?
+    var callback: TextToolOptionsView.Callback? = null
```
CodeChamp-SS, line 108: I think this `callback` var can be removed, as it isn't getting used anywhere.

CodeChamp-SS (commented):
Please look into these other minor changes :)

CodeChamp-SS (approved):

PR #1080, opened 2022-05-01
PAINTROID-146 set center tool with algorithm similar to scratch

davidmoling:
This PR is for the new set center feature in the transform tool.
https://jira.catrob.at/browse/PAINTROID-146

changes:

- added button for set center with the defined icon in transform tool
- refactored base class in order not to draw rectangle in transform tool
- fixed usage ToolOptionsViewController instead of ToolOptionsVisibilityController in multiple places
- added algorithm based on the idea that "Scratch" uses
- drawing a cursor in set center mode
- add history entry for set center
- added integration tests: one for a close center where and one for a further away center

### DIFF
--- a/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/tools/TransformToolIntegrationTest.kt
+++ b/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/tools/TransformToolIntegrationTest.kt
@@ -1334,6 +1334,44 @@ class TransformToolIntegrationTest {
         assertEquals(scale, perspective.scale, 0.0001f)
     }
 
+    @Test
+    fun testTransformToolSetCenterCloseCenter() {
+        drawPlus(layerModel.currentLayer!!.bitmap!!, initialWidth / 2)
+        onToolBarView()
+            .performSelectTool(ToolType.TRANSFORM)
+
+        runBlocking {
+            onTransformToolOptionsView().performSetCenterClick()
+        }
+        onDrawingSurfaceView()
+            .perform(UiInteractions.touchAt(DrawingSurfaceLocationProvider.BOTTOM_RIGHT_CLOSE_CENTER))
+        runBlocking {
+            TopBarViewInteraction.onTopBarView().performClickCheckmark()
+            delay(1500)
+        }
+        assertThat(toolSelectionBoxWidth, lessThan(initialWidth.toFloat()))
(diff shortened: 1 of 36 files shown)


Electronix1337:
Design needs to be adjusted, doesn't look good in the current state. I'd say you try to stick to the template of the ticket.

How it should look like:



Current look:

davidmoling:
Hi,

this design is on purpose, it give more interaction room since the elements are larger on purpose, it has more specific text (auto) and allows the edge case of e.g. 10000 pixels, which could happen. This was already merged in before in that way before the revert, so I'd suggest to keep it since its probably a better solution.

When do you get the nothging to resize? Steps pls

I would say that you should get that message only if you have not drawn anything yet. If that is not the case, I can correct that.

cheers

David

Electronix1337:
> Hi,
> 
> this design is on purpose, it give more interaction room since the elements are larger, it has more specific text (auto) and allows the edge case of e.g. 10000 pixels, which could happen. This was already merged in before in that way before the revert, so I'd suggest to keep it since its probably a better solution.
> 

Sorry I have to disagree, the only edge case I could think of is if a user wants to make a bitmap of 1 x 10000 but they are not even able to put a 5 digit number in there as our maximum resolution is bound to the metrics of the display (At least none of my emulators have such a high resolution, best I could get was a pixel XL where the maximum height was ~9500). There might be some tablets or phones out there which have such a high resolution that a 5 digit number would be possible but I don't think we should take that into consideration or rather make space for a 5 digit number but still fit everything in one line. As it's a user interface matter, I'm going to ask Thorsten to take a look at the design, but I'm quite certain that the current state is not the desired outcome.

> When do you get the nothging to resize? Steps pls
> 



> I would say that you should get that message only if you have not drawn anything yet. If that is not the case, I can correct that.
> 

This does not make sense to me, as the set center tool does not change the size of the bitmap if nothing was drawn yet or does it? We do have a toast "maximum image resolution reached" which is displayed when user wants to manually resize with the transform tool but the resulting bitmap would be larger than our max image resolution which is set to display width*height*4. I think it would be a good idea to either show a toast that the resulting bitmap would be bigger than the max image resolution or if the bitmap would exceed max image resolution, set it to the max image resolution, hope it's clear what I mean. 

> cheers
> 
> David

cheers 

Michael

Electronix1337 (approved):

wslany:
Thanks!

PR #1091, opened 2022-05-11
PAINTROID-283: add opacity for layers

StanglTobias:
added opacity for each layer. 
repaced transparentbitmap in layer by a simple flag.
added new layeropacity command

### DIFF
--- a/Paintroid/src/main/java/org/catrobat/paintroid/FileIO.kt
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/FileIO.kt
@@ -39,6 +39,7 @@ import androidx.annotation.RequiresApi
 import androidx.core.content.FileProvider
 import androidx.exifinterface.media.ExifInterface
 import id.zelory.compressor.Compressor
+import org.catrobat.paintroid.command.serialization.CommandSerializer
 import org.catrobat.paintroid.common.CATROBAT_IMAGE_ENDING
 import org.catrobat.paintroid.common.Constants.DOWNLOADS_DIRECTORY
 import org.catrobat.paintroid.common.Constants.PICTURES_DIRECTORY
@@ -52,7 +53,6 @@ import org.catrobat.paintroid.contract.MainActivityContracts
 import org.catrobat.paintroid.iotasks.BitmapReturnValue
 import org.catrobat.paintroid.iotasks.WorkspaceReturnValue
 import org.catrobat.paintroid.presenter.MainActivityPresenter
-import org.catrobat.paintroid.tools.Workspace
 import java.io.File
 import java.io.FileInputStream
 import java.io.FileOutputStream
@@ -547,9 +547,10 @@ object FileIO {
             inJustDecodeBounds = false
         }
         val scaling = hasEnoughMemory(resolver, bitmapUri, context)
+        val bitmap = enableAlpha(decodeBitmapFromUri(resolver, bitmapUri, options, context))
         return BitmapReturnValue(
             null,
-            enableAlpha(decodeBitmapFromUri(resolver, bitmapUri, options, context)),
+            bitmap,
             scaling
         )
     }
@@ -565,9 +566,11 @@ object FileIO {
             inJustDecodeBounds = false
             inSampleSize = getScaleFactor(resolver, bitmapUri, context)
         }
+
+        val bitmap = enableAlpha(decodeBitmapFromUri(resolver, bitmapUri, options, context))
         return BitmapReturnValue(
             null,
-            enableAlpha(decodeBitmapFromUri(resolver, bitmapUri, options, context)),
+            bitmap,
             false
         )
     }
@@ -585,14 +588,14 @@ object FileIO {
         return bitmap
     }
 
-    fun saveTemporaryPictureFile(internalMemoryPath: File, workspace: Workspace) {
+    fun saveTemporaryPictureFile(internalMemoryPath: File, commandSerializer: CommandSerializer) {
         val newFileName = "${TEMP_IMAGE_NAME}1.$CATROBAT_IMAGE_ENDING"
         val tempPath = File(internalMemoryPath, TEMP_IMAGE_DIRECTORY_NAME)
         try {
             tempPath.mkdirs()
 
             val stream = FileOutputStream("$tempPath/$newFileName")
-            workspace.getCommandSerializationHelper().writeToInternalMemory(stream)
+            commandSerializer.writeToInternalMemory(stream)
             temporaryFilePath = TEMP_IMAGE_TEMP_PATH
         } catch (e: IOException) {
             Log.e("Cannot write", "Can't write to stream", e)
@@ -635,12 +638,12 @@ object FileIO {
         temporaryFilePath = TEMP_IMAGE_PATH
     }
 
-    fun openTemporaryPictureFile(workspace: Workspace): WorkspaceReturnValue? {
+    fun openTemporaryPictureFile(commandSerializer: CommandSerializer): WorkspaceReturnValue? {
         var workspaceReturnValue: WorkspaceReturnValue? = null
         if (temporaryFilePath != null) {
             try {
                 val stream = FileInputStream(temporaryFilePath)
-                workspaceReturnValue = workspace.getCommandSerializationHelper().readFromInternalMemory(stream)
+                workspaceReturnValue = commandSerializer.readFromInternalMemory(stream)
             } catch (e: IOException) {
                 Log.e("Cannot read", "Can't read from stream", e)
             }
--- a/Paintroid/src/main/java/org/catrobat/paintroid/contract/LayerContracts.kt
+++ b/Paintroid/src/main/java/org/catrobat/paintroid/contract/LayerContracts.kt
@@ -37,11 +37,9 @@ interface LayerContracts {
         val layerCount: Int
         val presenter: Presenter
 
-        fun onBindLayerViewHolderAtPosition(
-            position: Int,
-            viewHolder: LayerViewHolder,
-            isOpen: Boolean
-        )
+        fun onSelectedLayerInvisible()
+
+        fun onSelectedLayerVisible()
 
         fun getListItemDragHandler(): ListItemDragHandler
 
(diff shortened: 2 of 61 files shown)


Paintroid/src/main/java/org/catrobat/paintroid/iotasks/OpenRasterFileFormatConversion.java
```diff
@@ -186,9 +198,9 @@ private static byte[] getXmlStack(List<Bitmap> bitmapList) {
 			Attr attr2 = doc.createAttribute("w");
 			Attr attr3 = doc.createAttribute("h");
 
-			attr1.setValue("0.0.1");
-			attr2.setValue(String.valueOf(bitmapList.get(0).getWidth()));
-			attr3.setValue(String.valueOf(bitmapList.get(0).getHeight()));
+			attr1.setValue(ORA_VERSION);
```
CodeChamp-SS, line 201: nice :)

Paintroid/src/main/java/org/catrobat/paintroid/model/Layer.kt
```diff
@@ -19,25 +19,16 @@
 package org.catrobat.paintroid.model
 
 import android.graphics.Bitmap
-import android.graphics.Color
 import org.catrobat.paintroid.contract.LayerContracts
 
-open class Layer(override var bitmap: Bitmap?) : LayerContracts.Layer {
-    override var transparentBitmap: Bitmap? = null
+open class Layer(override var bitmap: Bitmap) : LayerContracts.Layer {
```
CodeChamp-SS, line 24: the `bitmap` can be nullable, for eg. check `OpenRasterFileFormatConversion` line 276

Paintroid/src/main/java/org/catrobat/paintroid/model/Layer.kt
```diff
@@ -19,25 +19,16 @@
...
-        val tmpBitmap = transparentBitmap?.apply { copy(config, isMutable) }
-        transparentBitmap = bitmap
-        bitmap = tmpBitmap
-        if (isUnhide) {
-            transparentBitmap?.eraseColor(Color.TRANSPARENT)
-        }
+    companion object {
+        const val MAX_LAYER_OPACITY_PERCENTAGE = 100
+        const val MAX_LAYER_OPACITY_VALUE = 255
     }
```
CodeChamp-SS, line 33: you can make them top-level constants

Paintroid/src/main/java/org/catrobat/paintroid/model/LayerModel.kt
```diff
@@ -63,30 +62,25 @@ class LayerModel : LayerContracts.Model {
...
+                }
+                canvas?.drawBitmap(layer.bitmap, 0f, 0f, alphaPaint)
             }
-            return bitmapList
         }
+
+        return bitmap
     }
+
+    override fun getBitmapListOfAllLayers(): List<Bitmap?> = layers.map { it.bitmap }
```
CodeChamp-SS, line 85: nice 👍🏻

Paintroid/src/main/res/layout/pocketpaint_item_layer.xml
```diff
@@ -18,20 +19,18 @@
  -->
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:id="@+id/pocketpaint_item_layer_background"
-    android:layout_width="150dp"
-    android:layout_height="wrap_content"
+    android:layout_width="200dp"
+    android:layout_height="170dp"
```
CodeChamp-SS, line 23: sure about fixing size?

Paintroid/src/main/java/org/catrobat/paintroid/FileIO.kt
```diff
@@ -548,9 +548,10 @@ object FileIO {
             inJustDecodeBounds = false
         }
         val scaling = hasEnoughMemory(resolver, bitmapUri, context)
+        val bitmap = enableAlpha(decodeBitmapFromUri(resolver, bitmapUri, options, context))!!
         return BitmapReturnValue(
             null,
-            enableAlpha(decodeBitmapFromUri(resolver, bitmapUri, options, context)),
```
CodeChamp-SS, line 553: I think you can keep this as it is, any specific reason for the separate variable declaration?

CodeChamp-SS (changes requested):
These are some of the changes I could figure out, I'll mention if I find something more.
Could you please fix them? Thanks!

Paintroid/src/main/java/org/catrobat/paintroid/contract/LayerContracts.kt
```diff
@@ -109,11 +108,11 @@ interface LayerContracts {
     }
 
     interface Layer {
-        var bitmap: Bitmap?
-        var transparentBitmap: Bitmap?
+        var bitmap: Bitmap
```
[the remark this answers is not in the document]
StanglTobias: I am quite sure that it cannot be `null`. `DefaultCommandManager` line 80 does not state that the bitmap in `Layer` can `null`, just that the `Layer` itself can be `null`.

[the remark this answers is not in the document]
CodeChamp-SS: I mean if the layer is null, then `currentLayer?.bitmap` also becomes null

Paintroid/src/main/res/layout/pocketpaint_item_layer.xml
StanglTobias: same as above. The height is necassary because we support to display the preview with the correct aspect ratio. with wrap content this would also compress the opacity seekbar dramatically. Also a fixed height won't cause any problems

Paintroid/src/main/res/layout/pocketpaint_item_layer.xml
```diff
@@ -18,20 +19,18 @@
...
     android:orientation="horizontal"
-    android:paddingStart="8dp"
     android:paddingTop="10dp"
-    android:paddingEnd="8dp"
     android:paddingBottom="10dp"
     android:baselineAligned="false">
 
     <LinearLayout
-        android:layout_width="wrap_content"
+        android:layout_width="50dp"
```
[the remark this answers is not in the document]
CodeChamp-SS: hmm ok, in that case, we can let it be as you've implemented.

Paintroid/src/main/java/org/catrobat/paintroid/FileIO.kt
StanglTobias: i think this is more clean and readable.

Paintroid/src/main/java/org/catrobat/paintroid/contract/LayerContracts.kt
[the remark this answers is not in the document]
StanglTobias: what would be the purpose of making the bitmap nullable only because the layer can be null ?

Paintroid/src/main/java/org/catrobat/paintroid/ui/LayerAdapter.kt
```diff
@@ -160,4 +212,10 @@ class LayerAdapter(val presenter: LayerContracts.Presenter) : BaseAdapter(), Lay
 
         override fun setMergable() = layerBackground.setBackgroundResource(R.color.pocketpaint_color_merge_layer)
     }
+
+    companion object {
+        private val TAG = LayerAdapter::class.java.simpleName
+        private const val MIN_VAL = 0
+        private const val MAX_VAL = 100
```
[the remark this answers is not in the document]
CodeChamp-SS: Even when you make it top-level, it's still tied to `LayerAdapter`. You may refer https://stackoverflow.com/questions/61083630/when-should-i-prefer-top-level-const-vs-companion-object-const

Paintroid/src/main/java/org/catrobat/paintroid/FileIO.kt
CodeChamp-SS: hmm ok, then please remove the `!!` operators ;D

Paintroid/src/main/java/org/catrobat/paintroid/contract/LayerContracts.kt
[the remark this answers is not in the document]
CodeChamp-SS: sorry my bad, wrong reference 😅, the reason is similar to the one mentioned in `Layer.kt` comment

Paintroid/src/main/java/org/catrobat/paintroid/ui/LayerAdapter.kt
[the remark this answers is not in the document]
StanglTobias: Since nothing of that is stated in the Kotlin Coding Conventions I don't see one forum post being an offical recommendation. And in terms of readability I definitely prefer this option. But in terms of consistency I will make them top-level constants.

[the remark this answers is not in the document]
CodeChamp-SS: Thanks! I understand that there's no official convention for this, but Jetbrains prefer it, and this has been discussed in multiple articles/blogs, not only in one post I shared. Basically, the advantage of declaring top-level constants is less overhead.
https://www.egorand.dev/where-should-i-keep-my-constants-in-kotlin/
https://medium.com/kotlin-turf/do-you-need-a-companion-object-dc46fc6af7a8

Paintroid/src/main/java/org/catrobat/paintroid/model/Layer.kt
StanglTobias: for this reason `importOraFile` throws a nullpointerexception.

CodeChamp-SS: hmm ok, there's a null check implemented there, missed that.

FromG (approved):
Seems good to me. Tests should work. 
Images save the opacity.
To me the Slider seems a bit thin, however different opinions are needed.

CodeChamp-SS (approved):
LGTM!

wslany (changes requested):
Please eliminate the spurious rescaling after an opacity change in a layer using the soft keyboard on the number field, as discussed on Slack.

Electronix1337 (approved):
Overall LGTM, the only thing I am not quite sure is why is the color slightly different when i.e. setting a layer filled with black (#FF000000) and setting the opacity of the layer to 50% results in the color #7F000000 does not exactly result in the same color as setting black (#FF000000) in the colorpicker and then set alpha to 50%, the resulting color is #82000000. The difference is only every so slightly but still should not be the case imo.

ThorstenBandel:
LGTM!

PR #1094, opened 2022-05-16
PAINTROID-209 .ora file not correctly created on Samsung devices with Android 6.0 - 9

Electronix1337:
After this ticket the .ora and .catrobat-image files will be stored in the "Downloads" folder for all API's. For API < 29 I had to add the download manager which enables the media file scanner to detect and properly show the files in the file explorer. The user is also able to overwrite those files and also share/open with Paintroid. In order to make it work for API < 29 I also had to add the stored files as a downloaded file, which will be shown as a notification -->
image

There is also a boolean parameter "ShowNotification" but if I set this to false, the file saving does not work anymore.

The problem lies in the fact that for some devices/API's the "Downloads" folder will only show files if the files are downloaded, that's why I have to mimic a download even though the file is "just" saved. 

I've tried multiple devices/API's and was always able to see/overwrite/open all .ora and .catrobat-image files.

One thing which I was not able to fix is: If user stores .ora/.catrobat-image files we now have to store the Uri of the file as key of a shared preference and the download manager id as the value to properly remove the file if it gets overwritten. But if the user stores .catrobat-image/.ora files and the uninstalls the app and then again tries to save a .catrobat-image/.ora file with an already existing .catrobat-image/.ora file name, the overwrite dialog shows up, but the download manager is not able to remove the id as it was stored in the shared preferences, there will then be two files with the exact same filename. It's a neat edge case but I'm not sure if that's fine as I don't see this case happen too often.

This ticket might also fix: https://jira.catrob.at/browse/PAINTROID-348 as I was not able to find a device/API combination, where the .catrobat-image files were not loadable/openable 


FromG (approved):
ok

PR #1106, opened 2022-05-27
PAINTROID-419: "Vanishing" UI controls

FromG:
hide unhide tool options with animation.

 if the tool has top and bottom specific layouts. they will slide up and down.
 tools with only bottom layouts slide down.
 -added methods in every tool affected by the animation
 -also added methods in the default option views.

 -changed Drawingsurface that the drawing surface will begin at the top of the screen instead from under the topnavigationLayout and will end at the bottom of the screen instead of the navigationbottomlayout.

-changed headers of worked on files.

-transparency from top and bottom main navigation set to 15% when drawing.

-removes scrollingviewintegrationtests

-removes the auto scrolling feature. when you draw the drawingsurface will not scroll with the current direction the user draws.

-also hides the android status and navigation bar.

-remove autoscrolltasktests

-removed non full screen mode. the user will now always be in the fullscreen perspective.

-does not reset scale when you scale out and then draw on the zoomed out image.

-only resets scale when you switch modes.

https://jira.catrob.at/projects/PAINTROID/issues/PAINTROID-461?filter=allopenissues

- [ ] Post a message in the *#paintroid* [Slack channel](https://catrobat.slack.com) and ask for a code reviewer


Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/ScrollingViewIntegrationTest.java
```diff
@@ -281,24 +289,24 @@ public void dragAndCheckIfCanvasHasMovedInXAndY(PointF fromPoint, PointF toPoint
...
+//		int resourceId = mainActivity.getResources().getIdentifier("status_bar_height", "dimen", "android");
+//		if (resourceId > 0) {
+//			statusBarHeight = mainActivity.getResources().getDimensionPixelSize(resourceId);
+//		}
+
+//		int actionBarHeight;
+//		final TypedArray styledAttributes = mainActivity.getTheme().obtainStyledAttributes(
+//				new int[] {android.R.attr.actionBarSize}
+//		);
+//		actionBarHeight = (int) styledAttributes.getDimension(0, 0);
```
Electronix1337, line 302: As you already ignored the tests, I would suggest to comment in all lines so the one who does the fixing of the tests knows which lines were in the test case initially.

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/ScrollingViewIntegrationTest.java
```diff
@@ -129,11 +134,11 @@ public void testScrollingViewCursorTool() {
...
-		int statusBarHeight = 0;
-		int resourceId = mainActivity.getResources().getIdentifier("status_bar_height", "dimen", "android");
-		if (resourceId > 0) {
-			statusBarHeight = mainActivity.getResources().getDimensionPixelSize(resourceId);
-		}
+//		int statusBarHeight = 0;
+//		int resourceId = mainActivity.getResources().getIdentifier("status_bar_height", "dimen", "android");
+//		if (resourceId > 0) {
+//			statusBarHeight = mainActivity.getResources().getDimensionPixelSize(resourceId);
+//		}
```
Electronix1337, line 141: As you already ignored the tests, I would suggest to comment in all lines so the one who does the fixing of the tests knows which lines were in the test case initially.

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/ScrollingViewIntegrationTest.java
```diff
@@ -121,6 +124,8 @@ public void testScrollingViewDrawTool() {
 		longpressOnPointAndCheckIfCanvasPointHasChangedInXAndY(topRight);
 	}
 
+	// Ignore since Paintroid 419, needs to be rewritten.
```
Electronix1337, line 127: You can delete this comment and put the content in the ignore statement.

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/ScrollingViewIntegrationTest.java
```diff
@@ -121,6 +124,8 @@ public void testScrollingViewDrawTool() {
...
+	@Ignore
```
Electronix1337, line 128: You can write the line like this: @Ignore("since Paintroid 419, needs to be rewritten."), so the reasoning is in the ignore line and we also don't get another warning, that a test was ignored without reasoning.

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/ScrollingViewIntegrationTest.java
```diff
@@ -66,6 +67,8 @@ public void setUp() {
 		mainActivity = launchActivityRule.getActivity();
 	}
 
+	// Ignore since Paintroid 419, needs to be rewritten.
+	@Ignore
```
Electronix1337, line 71: You can write the line like this: @Ignore("since Paintroid 419, needs to be rewritten."), so the reasoning is in the ignore line and we also don't get another warning, that a test was ignored without reasoning.

Paintroid/src/androidTest/java/org/catrobat/paintroid/test/espresso/ScrollingViewIntegrationTest.java
```diff
@@ -66,6 +67,8 @@ public void setUp() {
 		mainActivity = launchActivityRule.getActivity();
 	}
 
+	// Ignore since Paintroid 419, needs to be rewritten.
```
Electronix1337, line 70: You can delete this comment and put the content in the ignore statement.

Paintroid/src/main/res/layout/activity_pocketpaint_main.xml
```diff
@@ -78,6 +78,7 @@
                 android:id="@+id/pocketpaint_main_bottom_navigation"
                 android:layout_width="match_parent"
                 android:layout_height="@dimen/pocketpaint_main_bottom_bar_height"
+                android:alpha="0.80"
```
Electronix1337, line 81: Might be 0.85 as the transparency shall be 15% if I'm not mistaken.

Paintroid/src/main/res/layout/pocketpaint_layout_top_bar.xml
```diff
@@ -23,6 +23,7 @@
     android:id="@+id/pocketpaint_layout_top_bar"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
+    android:alpha="0.80"
```
Electronix1337, line 26: Might be 0.85 as the transparency shall be 15% if I'm not mistaken.

Electronix1337 (changes requested):
Overall implementation LGTM and also seems pretty smooth. The only thing I'm not too sure if the vanishing shall also be applied for the smudge tool as most of the time the user has to swipe multiple times and then the vanishing actually is rather annoying than helpful, could only be on my side but wanted to mention it.

Paintroid/src/main/res/layout/pocketpaint_layout_top_bar.xml
FromG: this was discussed with wolfgang slany. we came to the conclusion that 0.8 is sufficient

Electronix1337 (approved):
Implementation LGTM, works as expected. Please fix the static warnings.

wslany (changes requested):
Thanks! 

- Please fix https://jenkins.catrob.at/job/Paintroid/job/PR-1106/28/analysis/
- Please rebase, this should fix the missing test executions: https://jenkins.catrob.at/job/Paintroid/job/PR-1106/test_results_analyzer/
- In a situation where the user makes a finger-stroke that leaves the screen, the top and bottom bars etc are not coming back --- it is as if the screen was still touched, even though it is not. Maybe the "end of touch" is then not correctly recorded in such a case?
- The canvas in the back's size and position on the screen should be shown in the same way as in the "fullscreen" mode, i.e., without the notification bar at the top of the phone. I believe that some of the issues described below will be automatically solved by this.
- Unexpected rescaling: in the default mode (with the vanishing top and bottom bars), do a zoom out so that the canvas becomes small. Then in the overhead menu, select the fullscreen mode. Observe that the canvas is unchanged, small (this is _correct_). Touch the screen and observe that then the canvas incorrectly is maximized (it should remain the same --- no rescaling nor any repositioning shall be done). The size and position of the canvas shall not change.
- Another unexpected rescaling: in the fullscreen mode (from the overflow menu), zoom out so that the canvas becomes small. Then tap on the Android "back" key to return to the default mode (with the vanishing bars). Observe that the canvas is unchanged, small (this is _correct_, it should remain small). Then touch the screen and observe that _incorrectly_ the canvas is suddenly maximized at that time. It should remain small, as shown before. The size and position of the canvas shall not change.
- Please rename the text on the "Fullscreen" menu entry in the overflow menu to "Hide buttons". 
- Currently, when in the default mode with the vanishing bars, and you slowly draw a line from the center to one of the borders, at the last moment the whole canvas is moved into the other direction on the screen. This is an old feature that was well intended then, but actually is not very useful and usually more hindering than useful. Please eliminate it in all tools and modes. The user can always actively achieve the same by moving the canvas on purpose.

Electronix1337 (changes requested):
Overall implementation LGTM but there are some issues which I want to at least address and let Wolfgang decide whether these are acceptable or not.

Zooming in and out seems to not work 100% correct, as after each zoom, the first click the user makes seems to be ignored and doesn't change anything on the canvas, the second click then shows the bars, which were vanished, again.

LineTool:
- When using the line tool and drawing a line, then clicking on plus to extend the line, the top bar vanishes when clicking on the plus button, which should not happen, as this is not happening, when user clicks undo, redo and checkmark, which are all also in the top bar but when clicked, the bar does not shortly vanish.

Overall single clicking with any tool does not seem to work as i.e. with the brush tool no dot is drawn and when single clicking and keeping the finger on the screen for a short amount of time, something is drawn but not a single dot but rather a small short line. This most likely happens because of vanishing the bars, which also slightly moves the canvas and therefore a small line rather than a dot is recognized. This is actually not really acceptable as users are not able to make small changes with this bug.

SmudgeTool:
- Smudge tool seems to be not really usable with this feature, at least when I'm trying to smudge a small portion of the display just for a bit, the handling of the clicks are to vague to make proper changes. One option would be to disable this feature for some tools, where the constant switch between vanishing and displaying is rather disturbing than helpful, like in this tool. But this should also be discussed in further detail.

Also if I understood Wolfgang correctly, the canvas shall not be maximized when clicking "hide buttons" in the settings menu. This is now the case, as well as when exiting the "hide buttons" mode.

It seems like the test: testTransformToolSetCenterFarCenter() fails since this PR, so please try to fix this before I can approve changes.

ThorstenBandel:
Finished in https://github.com/Catrobat/Paintroid/pull/1180

PR #1136, opened 2022-08-03
PAINTROID-420 Support more than four layers

Electronix1337:
I first implemented this ticket with the existing Listview and after finishing the implementation I experienced that performance when dragging and scrolling is too bad to be acceptable. Therefore I researched for a good alternative and found the Recyclerview which is a bit more work to implement and make it work but performance is better and overall Recyclerview seems more flexible. I then implemented the recyclerview and after that still experienced dropping frames which was caused by calling notifyDataSetChanged() too often, which is too generic for the recyclerview. Also changed the layout of the layermenu from a relative layout to a contraint layout which also improves performance. Changed number of MAX_LAYERS to 100. Also implemented a memory check, to verify that there is enough heap space to allocate new layer/bitmap. When user drags layer to bottom or top of the list, the list will automatically scroll down/up respectively and also visibly change with the last/first visible layer. 

Also implemented two tests, one to check if 100 layers can be added, if there is enough space, and also assert that "add layer" button will be disabled and one test to see if 10 layers can be added and that the layer count is correct.

For the sake of testing I tried adding a 2MB picture to each of the 100 layers and saving it as a .ora and .catrobat-image to see performance of saving and also if it works in general. The resulting file was 188 MB large and the process of saving took around 1-2 minutes. I shortly investigated if this could be improved by any means but did not find anything, as we have to save each bitmap separately and we are already compressing it with a PNG format. Also the loading of such a large file takes quite some time but this could be improved in another ticket maybe.

https://jira.catrob.at/browse/PAINTROID-420


AneiMakovec (approved):
LGTM. More than 4 layers can be added, scrolled through, grabbed and dragged around, as expected. Tests are also ok.

wslany (approved):
Thanks!!

PR #1180, opened 2023-01-03
PAINTROID-419 "Vanishing" UI controls

Lenkomotive:
[PAINTROID-419](https://jira.catrob.at/browse/PAINTROID-419)

Last Comments made in https://github.com/Catrobat/Paintroid/pull/1106 where i picked up from:

1. Zooming in and out seems to not work 100% correct, as after each zoom, the first click the user makes seems to be ignored and doesn't change anything on the canvas, the second click then shows the bars, which were vanished, again.

2. When using the line tool and drawing a line, then clicking on plus to extend the line, the top bar vanishes when clicking on the plus button, which should not happen, as this is not happening, when user clicks undo, redo and checkmark, which are all also in the top bar but when clicked, the bar does not shortly vanish.

3. Overall single clicking with any tool does not seem to work as i.e. with the brush tool no dot is drawn and when single clicking and keeping the finger on the screen for a short amount of time, something is drawn but not a single dot but rather a small short line. This most likely happens because of vanishing the bars, which also slightly moves the canvas and therefore a small line rather than a dot is recognized. This is actually not really acceptable as users are not able to make small changes with this bug.

4. Smudge tool seems to be not really usable with this feature, at least when I'm trying to smudge a small portion of the display just for a bit, the handling of the clicks are to vague to make proper changes. One option would be to disable this feature for some tools, where the constant switch between vanishing and displaying is rather disturbing than helpful, like in this tool. But this should also be discussed in further detail.

5. Also if I understood Wolfgang correctly, the canvas shall not be maximized when clicking "hide buttons" in the settings menu. This is now the case, as well as when exiting the "hide buttons" mode.

6. It seems like the test: testTransformToolSetCenterFarCenter() fails since this PR, so please try to fix this before I can approve changes.

New in this PR:

1.+ 3. Every touch event caused the statusbar to hide. Respectivly the drawingSurface changed its position. Now the app hides the statusbar by default. 
positive: bigger drawing surface , all errors from 1 and 3 are solved, no unnecessary overhead for calculating the correct recentering of the drawing surface. 
negative: user can see status bar only when swiping down

2.Wrapped the line-handling logic in a new method, so no animations are triggered when clicking the button

4.deactivated animations in smudge tool
-> redesign of smudge tool menu needed because it uses too much space

5.removed resetScaleAndTranslation() call from hideButtons calls
 
6.This tests fails locally also in delvelop branch. 

7.The animations were not working correctly in landscape mode, this was also fixed.

+ adjusted testcases (no autoscrolltask anymore)


Lenkomotive:
Last Comments made in https://github.com/Catrobat/Paintroid/pull/1106 where i picked up from:

1.
Zooming in and out seems to not work 100% correct, as after each zoom, the first click the user makes seems to be ignored and doesn't change anything on the canvas, the second click then shows the bars, which were vanished, again.

2.
When using the line tool and drawing a line, then clicking on plus to extend the line, the top bar vanishes when clicking on the plus button, which should not happen, as this is not happening, when user clicks undo, redo and checkmark, which are all also in the top bar but when clicked, the bar does not shortly vanish.

3.
Overall single clicking with any tool does not seem to work as i.e. with the brush tool no dot is drawn and when single clicking and keeping the finger on the screen for a short amount of time, something is drawn but not a single dot but rather a small short line. This most likely happens because of vanishing the bars, which also slightly moves the canvas and therefore a small line rather than a dot is recognized. This is actually not really acceptable as users are not able to make small changes with this bug.

4.
Smudge tool seems to be not really usable with this feature, at least when I'm trying to smudge a small portion of the display just for a bit, the handling of the clicks are to vague to make proper changes. One option would be to disable this feature for some tools, where the constant switch between vanishing and displaying is rather disturbing than helpful, like in this tool. But this should also be discussed in further detail.

5.
Also if I understood Wolfgang correctly, the canvas shall not be maximized when clicking "hide buttons" in the settings menu. This is now the case, as well as when exiting the "hide buttons" mode.

6.
It seems like the test: testTransformToolSetCenterFarCenter() fails since this PR, so please try to fix this before I can approve changes.

New in this PR:

1.+ 3.

Every touch event caused the statusbar to hide. Respectivly the drawingSurface changed its position. Now the app hides the statusbar by default. 
positive: bigger drawing surface , all errors from 1 and 3 are solved, no unnecessary overhead for calculating the correct recentering of the drawing surface. 
negative: user can see status bar only when swiping down

2. 
Wrapped the line-handling logic in a new method, so no animations are triggered when clicking the button

4.
deactivated animations in smudge tool
-> redesign of smudge tool menu needed 

5. 
removed resetScaleAndTranslation() call from hideButtons calls
 
6.
The animations were not working correctly in landscape mode, this was also fixed.

SebastianGrief:
Some things I noticed:

- The tool menu doesn't disappear on draw, which is the same behaviour as before, but kind of wierd now with the bars disappearing. Also not the best behaviour in the first place imho. 
Reproduce: Pick a tool -> reopen the tools menu -> draw 
Maybe @juliajulie95 can comment on whether we want to open a new ticket for that.

- The reappearing isn't working as intended (I think?). For me they do not slide in but pop up and slide to their end position. 
So the animation (slideUp, slideDown) should depend on whether the topBar and bottomNavigation are appearing or disappearing.

SebastianGrief (changes requested):
See comment

juliajulie95:
Please fix the tool window not disappearing (@SebastianGrief mentioned it above)

Lenkomotive:
Both findings are fixed now. @SebastianGrief @juliajulie95

SebastianGrief (approved):

SebastianGrief (approved):
👍

ThorstenBandel:
There are more tests than normally failing. The following tests do not fail on develop. Could you take a look at those?

- testTransformToolSetCenterFarCenter
- testShapeToolBoxGetsPlacedCorrectWhenZoomedIn 
- testTextToolBoxIsPlacedCorrectlyWhenZoomedIn
- testTextToolDoesNotResetPerspectiveScale

Lenkomotive:
> 
testTransformToolSetCenterFarCenter is fixed now and should not fail anymore.  other three are also failing on develop sometimes

ThorstenBandel (approved):
Thanks!

PR #1302, opened 2023-08-16
PAINTROID-648 Dynamic animation of top and bottom bar

Lenkomotive:
Ticket: [PAINTROID-648](https://jira.catrob.at/browse/PAINTROID-648)

Added the threshold calculation to the handleMove method of Tool. Now it will only animate if the touchevent is within the first or last quarter of the devices display.

### DIFF
--- a/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/junit/tools/LineToolTest.kt
+++ b/Paintroid/src/androidTest/java/org/catrobat/paintroid/test/junit/tools/LineToolTest.kt
@@ -191,7 +191,7 @@ class LineToolTest {
         val tap1 = PointF(7f, 7f)
         Mockito.`when`(toolOptionsViewController.isVisible).thenReturn(true)
         Mockito.`when`(viewMock.visibility).thenReturn(View.VISIBLE)
-        tool.handleMove(tap1)
+        tool.handleMove(tap1, true)
 
         Mockito.verify(toolOptionsViewController).slideUp(viewMock,
                                                           willHide = true,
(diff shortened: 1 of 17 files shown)


bakicelebi (changes requested):
Please also animate top and bottom bar when using pipette tool.

bakicelebi (approved):
Okay, makes sense :+1:

JulianJautz (changes requested):
Dynamic animation works fine. 
But when using the Smudge tool, the bars are always there. Please add the animation for the Smudge tool.
Regarding, Shape, Clipboard and Text I assume you did not change them on purpose?

Lenkomotive:
> Dynamic animation works fine. But when using the Smudge tool, the bars are always there. Please add the animation for the Smudge tool. Regarding, Shape, Clipboard and Text I assume you did not change them on purpose?

**Shape, Clipboard and Text:** Yes, thats correct. I left them on purpose since the shapes are usually already big relative to screen size and therefore going into fullscreen mode immediately makes sense.

**Smudge**: I have chosen to omit the animation for the smudge tool. Given its frequent touch events during use, the animation could become disturbing for the user. Also, the line drawn by the smudge tool is typically short and I believe the desired area on the canvas is often visible to the user so that going into fullscreen mode is not that necessary.

JulianJautz:
> > Dynamic animation works fine. But when using the Smudge tool, the bars are always there. Please add the animation for the Smudge tool. Regarding, Shape, Clipboard and Text I assume you did not change them on purpose?
> 
> **Shape, Clipboard and Text:** Yes, thats correct. I left them on purpose since the shapes are usually already big relative to screen size and therefore going into fullscreen mode immediately makes sense.
> 
> **Smudge**: I have chosen to omit the animation for the smudge tool. Given its frequent touch events during use, the animation could become disturbing for the user. Also, the line drawn by the smudge tool is typically short and I believe the desired area on the canvas is often visible to the user so that going into fullscreen mode is not that necessary.

Makes sense. :)

JulianJautz (approved):

JulianJautz (changes requested):
Just one thing: What do you think of adding a "false" default value for shouldAnimate in the  handleMove method in Tool.kt? Then we would not need to change all the test classes.

Lenkomotive:
> Just one thing: What do you think of adding a "false" default value for shouldAnimate in the handleMove method in Tool.kt? Then we would not need to change all the test classes.

good idea! Done 👍🏼

